PR #11, opened 2025-08-02
feat: Add Wrap To Input Window

vkhitrin:
Adding an option to wrap words inside the `opencode_input` window.


sudo-tee:
This is a great idea

Thanks for the contribution.

PR #15, opened 2025-08-25
fix: detect opencode jsonc config files

Bekaboo:
Opencode supports jsonc config files but this plugin is not detecting them: https://opencode.ai/docs/config/#format

### DIFF
--- a/lua/opencode/config_file.lua
+++ b/lua/opencode/config_file.lua
@@ -11,7 +11,9 @@ M.ConfigKeys = {
 
 local config_files_names = {
   'config.json',
+  'config.jsonc',
   'opencode.json',
+  'opencode.jsonc',
 }
 
 function M.setup()


Bekaboo:
This would require a JSONC decoder in nvim to decode config options in opencode JSONC config files...

sudo-tee:
I came to the same conclusion that it's harder than it looks. 

I should add a note in the README in the meantime.

Bekaboo:
> I came to the same conclusion that it's harder than it looks.
>
> I should add a note in the README in the meantime.

It makes sense.

Bekaboo:
It would be nice if opencode can open up an API for clients to get its parsed config structure as it already have a jsonc parser and must have parsed the jsonc config file on launch. This could benefit other clients without a json/jsonc parser too. I am considering open a PR in opencode repo

sudo-tee:
I think they already do if you start a headless server with `opencode serve`  and hit the {server_url}/config  you will get a parsed config.

However this plugin is not starting a headless server it's using the cli `opencode run` .  It would require re-architecting the entire plugin to use the headless server instead, and also it would means that opencode would spawn their lsp servers in parallel to nvim, it would not be a lightweight plugin anymore.

sudo-tee:
I found that Plenary has some utils https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/json.lua to deal with json with comments.

this plugin already uses plenary

Bekaboo:
> I found that Plenary has some utils https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/json.lua to deal with json with comments.
> 
> this plugin already uses plenary

Awesome! If you prefer to implement this yourself feel free to close this PR (I will be busy in the following weeks).

sudo-tee:
Since I implemented some server api, I am now able to fetch the config without being bothered by the format by calling the /config endpoint from an ephemeral opencode server. 

So this makes the PR not relevent anymore.

PR #16, opened 2025-08-25
fix: colorscheme/user cannot override output hlgroups

Bekaboo:
Previously, opencode resets the hlgroup every time it opens the output window. This PR uses `default=true` to allow colorscheme/user to override opencode hlgroups.


Bekaboo:
Also added highlight group `OpenCodeHint` for hinting messages in the input window and token info in the output window to avoid using `Comment` highlight directly. This would allow users to customize the color of the hinting messages without changing the global `Comment` highlight group.

sudo-tee:
Thanks for this fix.

PR #17, opened 2025-08-25
fix: set `winfixheight` and `winfixwidth` for input/output windows

Bekaboo:
Set `winfixheight` and `winfixwidth` for input/output windows so that `<C-w>=` or opening/closing normal split windows will not accidentally resize the opencode windows.


sudo-tee:
This is great

Thanks for the fix

PR #18, opened 2025-08-25
fix: footer 'invalid window id' error when closing the output window

Bekaboo:
Getting the following error on pressing `<C-w>c` in the output window:

image

This PR fixes this issue by validating output window before updating the footer window.


sudo-tee:
Thanks again, 

I really appreciate theses little fixes

PR #21, opened 2025-08-26
feat: allow disabling buf-local keymaps

Bekaboo:
Currently, each window action must have a corresponding keymap, and there is no way to disable buf-local keymaps. Buf-local keymaps should be disabled if it is assigned to `false`, which aligns with the global keymap config.

Example use case: I have `<Tab>` remapped to tabout, so I don't want it to be mapped to "toggle pane". I also prefer to switch between input/output panel directly with `<M-j/k>` (mapped to `<C-w>j/k`), so I'd rather not assign a keymap for the "toggle pane" action.

### DIFF
--- a/lua/opencode/keymap.lua
+++ b/lua/opencode/keymap.lua
@@ -18,12 +18,16 @@ function M.setup(keymap)
   end
 end
 
----@param lhs string The left-hand side of the mapping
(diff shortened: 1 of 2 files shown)


sudo-tee:
It makes a ton of sense

Thanks for this.

Issue #28, opened 2025-09-24
Error attaching file to message


rustybucket-cloud:
After sending a message with a project file attached, I got the following error:

```
Error executing vim.schedule lua callback: vim/_editor.lua:0: Invalid chunk: expected Array with 1 or 2 Strings
stack traceback:
        [C]: in function 'nvim_echo'
        vim/_editor.lua: in function 'notify'
        ...ocal/share/nvim/lazy/opencode.nvim/lua/opencode/core.lua:193: in function 'fn'
        ...ocal/share/nvim/lazy/opencode.nvim/lua/opencode/util.lua:270: in function <...ocal/share/nvim/lazy/opencode.nvim/lua/opencode/util.lua:269>
Press ENTER or type command to continue
``` 

After cloning the repo and patching the following function in `util.lua` like this:
```
    on_error = function(err)
      state.opencode_server_job = nil
      state.last_output = os.time()
      ui.render_output()
      local msg = err
      if type(err) == 'table' then
        msg = err.message or err.msg or err.error or vim.inspect(err)
      end
      vim.notify(tostring(msg), vim.log.levels.ERROR)
      -- vim.notify(err.message, vim.log.levels.ERROR)
      util.safe_call(opts.on_error, err)
    end,
```

I got the following error:
```
{
  data = {
    message = "Error: ENOENT: no such file or directory, scandir '/pages/reporting/dashboard/corpora-dashboard'"
  },
  name = "UnknownError"
}
```

I got the error to resolve after making the following changes:
```
diff --git a/lua/opencode/context.lua b/lua/opencode/context.lua
index 2263a26..d3bb6cf 100644
--- a/lua/opencode/context.lua
+++ b/lua/opencode/context.lua
@@ -96,8 +96,30 @@ function M.add_file(file)
     M.context.mentioned_files = {}
   end
 
+  if not file or file == '' then
+    vim.notify('No file path provided to add_file', vim.log.levels.WARN)
+    return
+  end
+
+  -- Normalize: if path starts with '/' but is inside current working dir, make it absolute via cwd
+  local cwd = vim.fn.getcwd()
+  if vim.startswith(file, '/') and not vim.startswith(file, cwd) then
+    local candidate = cwd .. file
+    if vim.fn.filereadable(candidate) == 1 then
+      file = candidate
+    end
+  end
+
+  -- Expand to absolute path to avoid mismatches
+  file = vim.fn.fnamemodify(file, ':p')
+
[...]

PR #29, opened 2025-09-25
fix: standardize relative path

rustybucket-cloud:
Fix for #28.

Inside of a node project with a `baseUrl` of `src`, opencode is looking for `/pages/...` when the file was at `src/pages/...`. 

Expanding the path to an absolute, normalized path with `fnamemodify` fixes this.


sudo-tee:
Thank you for the fix, 

I've tested it and it works perfectly

PR #35, opened 2025-09-29
fix: broken link in README

rustybucket-cloud:
https://opencode.ai/docs/user-commands/ goes to a 404 page.

I think it's supposed to be pointing to https://opencode.ai/docs/commands/


sudo-tee:
Thanks for the fix. 

They probably changed the url

PR #47, opened 2025-10-06
feat: implement basic permissions worflow

sudo-tee:
This PR is adding permission workflow similar to the opencode one where you can accept, accept all or deny a tool request.

The prompt is not blocking the main nvim ui but rather just display a message in the chat window 

Screenshot 2025-10-06 105302
Screenshot 2025-10-06 105253

This PR also replace plenary.curl because it was causing issues with the /event api where it was buffering some events and not sending them immediately (it was crucial for the the permission api) to have them ASAP.


av1155:
I tested the new permissions feature and it works! I like to have my permissions set to ask by default so that helps a lot. Thank you.

---

Screenshot 2025-10-07 at 12 06 29 AM

---

Screenshot 2025-10-07 at 12 06 50 AM

PR #48, opened 2025-10-07
feat: add modes to keymap configs

cameronr:
I was overriding input/output buffer local keymaps (particularly `next/prev_prompt_history` and `toggle_pane`)  using Filetype autocmds and figured it was probably better to just make them configurable.

Should be backwards compatible.

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -25,6 +25,7 @@ M.defaults = {
   default_global_keymaps = true,
   default_mode = 'build',
   keymap = {
+    -- Global keymaps (all use { 'n', 'v' } modes by default)
     global = {
       toggle = '<leader>og',
       open_input = '<leader>oi',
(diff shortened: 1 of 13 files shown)


lua/opencode/config.lua
```diff
@@ -175,6 +175,27 @@ function M.setup(opts)
...
   end
+
+  -- Handle backward compatibility for submit_insert
+  if M.values.keymap and M.values.keymap.window and M.values.keymap.window.submit_insert then
+    vim.notify(
+      'opencode.nvim: submit_insert keymap is deprecated. Use submit = { key, { "n", "i" } } instead.',
+      vim.log.levels.WARN
+    )
+    
+    -- If user has submit_insert but not submit, migrate submit_insert to submit with multi-mode
```
sudo-tee, line 186: This might cause issues. Since sumbit_insert can be a different keymap.  I have mine set to ctrl-s because I like having multiple lines in my prompt. But still use <Cr> in normal mode.

lua/opencode/ui/input_window.lua
```diff
@@ -179,31 +187,64 @@ function M.setup_keymaps(windows)
...
-  map(keymaps.mention, completion.trigger_completion(keymaps.mention), input_buf, 'i')
-  map(keymaps.slash_commands, completion.trigger_completion(keymaps.slash_commands), input_buf, 'i')
-  map(keymaps.mention_file, core.add_file_to_context, input_buf, 'i')
-
-  map(keymaps.prev_prompt_history, nav_history(keymaps.prev_prompt_history, 'prev'), input_buf, { 'n', 'i' })
-  map(keymaps.next_prompt_history, nav_history(keymaps.next_prompt_history, 'next'), input_buf, { 'n', 'i' })
-
-  map(keymaps.switch_mode, api.switch_to_next_mode, input_buf, { 'n', 'i' })
+  -- Define keymap handlers table - functions that need the parsed key get it via closure
+  local keymap_handlers = {
```
sudo-tee, line 191: My original idea of this would have to expose every mapping as api.the_function

It would simplify things and could be a simple loop to apply them

sudo-tee:
I can also help move stuff around to the api to simplify things

You can look at this branch where I added a corresponding api function for each keymap function

https://github.com/sudo-tee/opencode.nvim/tree/keymaps_cleanup

cameronr:
Ooh, that's an even better idea, both because it follows a more common convention and because it's more flexible (ie could define multiple keys for same action).

I'll take a look at it. Should I build off the keymaps_cleanup branch or just use it for reference?

sudo-tee:
You can build from the `keymaps_cleanup` or rebase it into your branch since the job of cleaning them up is done there.

The only tricky part would be the permissions keymap that are applied in a dynamic way (I merged them this morning)

https://github.com/sudo-tee/opencode.nvim/blob/keymaps_cleanup/lua/opencode/keymap.lua#L61-L61

But maybe always apply them and use `vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes(key, true, false, true), mode, true), in the correcponding api function if there is no permission request.`

I will do a test on my branch

UPDATE: it would be to complex to do that.  they will have to be treated differently..  (keep them as they are) instead of putting them in the new config setup...

Issue #49, opened 2025-10-07
Split placement


michaelcosj:
Hi, thanks for the awesome plugin!

I’m seeing an unexpected split placement behavior. When I have multiple splits open and trigger OpenCode, the new windows open immediately to the right of the current split (or to the left depending on the position set in config), rather than at the outermost edge. I would expect them to open as the rightmost (or leftmost) split.

A video for reference:

https://github.com/user-attachments/assets/a256aea6-54b6-4df7-ba35-578aed215626

sudo-tee:
Thank you very much for providing a PR to fix the issue.

I merged the PR so I will close this issue

PR #50, opened 2025-10-07
Consistent split placement

michaelcosj:
Addresses #49


sudo-tee:
Thanks for the PR. 

I also like it better that way

PR #52, opened 2025-10-08
feat(config): modern keymap format + mode support

cameronr:
Here's the updated PR to have global and window keymaps be key first.

I went back and forth on it, but I currently have window permissions split out into their own table and have them action first. (but if you want it the other way, I can change it so they're also key first).

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -14,6 +14,8 @@
 --- @field get fun(key: "providers"): OpencodeProviders
 --- @field get fun(key: "context"): OpencodeContextConfig
 --- @field get fun(key: "debug"): OpencodeDebugConfig
+--- @field get_key_for_function fun(scope: 'global'|'window', function_name: string): string|nil
+--- @field normalize_keymap fun(keymap_config: table): table
 
 local M = {} ---@type OpencodeConfigModule
 
@@ -26,51 +28,52 @@ M.defaults = {
   default_mode = 'build',
   keymap = {
     global = {
-      toggle = '<leader>og',
-      open_input = '<leader>oi',
-      open_input_new_session = '<leader>oI',
-      open_output = '<leader>oo',
-      toggle_focus = '<leader>ot',
-      close = '<leader>oq',
-      select_session = '<leader>os',
-      configure_provider = '<leader>op',
-      diff_open = '<leader>od',
-      diff_next = '<leader>o]',
-      diff_prev = '<leader>o[',
-      diff_close = '<leader>oc',
-      diff_revert_all_last_prompt = '<leader>ora',
-      diff_revert_this_last_prompt = '<leader>ort',
-      diff_revert_all = '<leader>orA',
-      diff_revert_this = '<leader>orT',
-      diff_restore_snapshot_file = '<leader>orr',
-      diff_restore_snapshot_all = '<leader>orR',
-      swap_position = '<leader>ox', -- Swap Opencode pane left/right
-      permission_accept = '<leader>opa',
-      permission_accept_all = '<leader>opA',
-      permission_deny = '<leader>opd',
+      ['<leader>og'] = 'toggle',
+      ['<leader>oi'] = 'open_input',
+      ['<leader>oI'] = 'open_input_new_session',
+      ['<leader>oo'] = 'open_output',
+      ['<leader>ot'] = 'toggle_focus',
+      ['<leader>oq'] = 'close',
+      ['<leader>os'] = 'select_session',
+      ['<leader>op'] = 'configure_provider',
+      ['<leader>od'] = 'diff_open',
+      ['<leader>o]'] = 'diff_next',
+      ['<leader>o['] = 'diff_prev',
+      ['<leader>oc'] = 'diff_close',
+      ['<leader>ora'] = 'diff_revert_all_last_prompt',
+      ['<leader>ort'] = 'diff_revert_this_last_prompt',
+      ['<leader>orA'] = 'diff_revert_all',
+      ['<leader>orT'] = 'diff_revert_this',
+      ['<leader>orr'] = 'diff_restore_snapshot_file',
+      ['<leader>orR'] = 'diff_restore_snapshot_all',
+      ['<leader>ox'] = 'swap_position',
+      ['<leader>opa'] = 'permission_accept',
+      ['<leader>opA'] = 'permission_accept_all',
+      ['<leader>opd'] = 'permission_deny',
     },
     window = {
-      submit = '<cr>',
-      submit_insert = '<cr>',
-      close = '<esc>',
-      stop = '<C-c>',
-      next_message = ']]',
-      prev_message = '[[',
-      mention_file = '~',
-      mention = '@',
-      slash_commands = '/',
-      toggle_pane = '<tab>',
-      prev_prompt_history = '<up>',
-      next_prompt_history = '<down>',
-      switch_mode = '<M-m>',
-      focus_input = '<C-i>',
-      select_child_session = '<leader>oS',
-      debug_message = '<leader>oD',
-      debug_output = '<leader>oO',
-      debug_session = '<leader>ods',
-      permission_accept = 'a',
-      permission_accept_all = 'A',
-      permission_deny = 'd',
+      ['<cr>'] = { 'submit_input_prompt', mode = { 'n', 'i' } },
+      ['<esc>'] = 'close',
+      ['<C-c>'] = 'stop',
+      [']]'] = 'next_message',
+      ['[['] = 'prev_message',
+      ['~'] = { 'mention_file', mode = 'i' },
+      ['@'] = { 'mention', mode = 'i' },
+      ['/'] = { 'slash_commands', mode = 'i' },
+      ['<tab>'] = { 'toggle_pane', mode = { 'n', 'i' } },
+      ['<up>'] = { 'prev_prompt_history', mode = { 'n', 'i' } },
+      ['<down>'] = { 'next_prompt_history', mode = { 'n', 'i' } },
+      ['<M-m>'] = 'switch_mode',
+      ['<C-i>'] = 'focus_input',
+      ['<leader>oS'] = 'select_child_session',
+      ['<leader>oD'] = 'debug_message',
+      ['<leader>oO'] = 'debug_output',
+      ['<leader>ods'] = 'debug_session',
+    },
+    permission = {
+      accept = 'a',
+      accept_all = 'A',
+      deny = 'd',
     },
   },
   ui = {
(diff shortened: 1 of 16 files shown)


lua/opencode/config.lua
```diff
@@ -26,51 +28,53 @@ M.defaults = {
...
+      ['<cr>'] = { 'submit_input_prompt', mode = { 'n', 'i' } },
+      ['<esc>'] = 'close',
+      ['<C-c>'] = 'stop',
+      [']]'] = 'next_message',
+      ['[['] = 'prev_message',
+      ['~'] = { 'mention_file', mode = 'i' },
+      ['@'] = { 'mention', mode = 'i' },
+      ['/'] = { 'slash_commands', mode = 'i' },
+      ['<tab>'] = { 'toggle_pane', mode = { 'n', 'i' } },
+      ['<up>'] = { 'prev_prompt_history', mode = { 'n', 'i' } },
```
sudo-tee, line 65: This could be

lua/opencode/config.lua
```diff
@@ -26,51 +28,53 @@ M.defaults = {
...
+      ['<down>'] = { 'next_prompt_history', mode = { 'n', 'i' } },
+      ['<M-m>'] = 'switch_mode',
+      ['<C-i>'] = 'focus_input',
```
sudo-tee: I had a bug here.. Mine was set to  `i` instead of `<C-i>` and I could not enter insert mode in the input window.  this keymap is only intended for the output window

lua/opencode/config.lua
```diff
@@ -26,51 +28,53 @@ M.defaults = {
...
+      ['<esc>'] = 'close',
+      ['<C-c>'] = 'stop',
+      [']]'] = 'next_message',
+      ['[['] = 'prev_message',
+      ['~'] = { 'mention_file', mode = 'i' },
+      ['@'] = { 'mention', mode = 'i' },
+      ['/'] = { 'slash_commands', mode = 'i' },
+      ['<tab>'] = { 'toggle_pane', mode = { 'n', 'i' } },
+      ['<up>'] = { 'prev_prompt_history', mode = { 'n', 'i' } },
+      ['<down>'] = { 'next_prompt_history', mode = { 'n', 'i' } },
```
sudo-tee, line 60: These 2 are only intended for the input window.

lua/opencode/config.lua
```diff
@@ -189,4 +250,44 @@ function M.get(key)
...
+  local config_data = M.get()
+
+  local keymap_config = config_data.keymap and config_data.keymap[scope]
+  if not keymap_config then
+    return nil
+  end
+
+  -- All configs are normalized after setup, so only handle new format
+  for key, config in pairs(keymap_config) do
+    if type(config) == 'string' then
```
sudo-tee, line 237: Maybe only support one format     ['the_mapping'] = {'fuction_name'  or function ()  }  instead of ['the_mapping'] = 'function_name' so you don't have to do this check everytime ?

lua/opencode/config.lua
```diff
@@ -163,6 +167,45 @@ M.defaults = {
 
 M.values = vim.deepcopy(M.defaults)
 
+--- Check if a keymap configuration uses the old format (internal use only)
+--- @param keymap_config table
+--- @return boolean
+local function is_old_format(keymap_config)
```
sudo-tee: With the updated config format you could just check for the `window` or `global` key under keymaps

lua/opencode/config.lua
```diff
@@ -26,51 +28,53 @@ M.defaults = {
...
+      ['<tab>'] = { 'toggle_pane', mode = { 'n', 'i' } },
+      ['<up>'] = { 'prev_prompt_history', mode = { 'n', 'i' } },
+      ['<down>'] = { 'next_prompt_history', mode = { 'n', 'i' } },
+      ['<M-m>'] = 'switch_mode',
+      ['<C-i>'] = 'focus_input',
+      ['<leader>oS'] = 'select_child_session',
+      ['<leader>oD'] = 'debug_message',
+      ['<leader>oO'] = 'debug_output',
+      ['<leader>ods'] = 'debug_session',
+    },
```
sudo-tee, line 66: I think we should separate input and output window keymaps.

```lua
keymaps = {
    editor = {
        
    },
    input_window = {
           ['<down>'] = { 'next_prompt_history', mode = { 'n', 'i' } },
           ...
    },
    output_window = {
          ['<C-i>'] = {'focus_input'}, -- only 1 format to simplify ?
           ...
    },
   global = {
        ...legacy
    },
    window = {
        -- legacy
    }
}

```

This should make the legacy detection easier at the expanse of duplicating some keymaps for flexibility

sudo-tee (commented):
Really great PR. 

I only added some small comments about the format and about splitting input_window and output_window keymaps  and using `editor` instead of `global`.  this should really simplify `old format detection` and  `keymap applying`. 

We could support custom function as well so users can plug whatever they want in there :) 

I know the format changed a little bit since the last PR.. But I want to be sure we make it right and future-proof and also make it easy to maintain.

But for the rest this is a really great PR and it tackle a really core aspect of the plugin in a elegant way :).

With this last round of fixes it will be ready to merge

Thanks for your effort on this. It's really appreciated


Here is a recap of the  updated format that I think we should go with

```lua
keymaps = {
    editor = {
         ['<leader>og'] = { 'toggle' },
    },
    input_window = {
           ['<down>'] = { 'next_prompt_history', mode = { 'n', 'i' } }, -- if you want to specify mode
           ['<M-u>'] = {function () 
                          ---- custom user code
             end, mode={'i'}}
           ...
    },
    output_window = {
          ['<C-i>'] = {'focus_input'}, -- only 1 format to simplify ?
           ...
    },
   global = {
        ...legacy
    },
    window = {
        -- legacy
    }
}
```

lua/opencode/keymap.lua
```diff
@@ -8,11 +8,33 @@ function M.setup(keymap)
...
+          modes = config_entry.mode
+        end
+      end
+    end
+
+    if func_name and api[func_name] then
+      vim.keymap.set(modes, key_binding, function()
+        api[func_name]()
+      end, { silent = false, desc = cmds[func_name] and cmds[func_name].desc })
     end
```
sudo-tee: With the new format it can become much simpler

```lua
function M.setup(keymap)
  local api = require('opencode.api')
  local cmds = api.commands
  local global = keymap.editor or {}

  for key_binding, config_entry in pairs(global) do
    local func_name = config_entry[1]
    local modes = config_entry.mode or { 'n', 'v' }  -- vim.keymap.set already support string or string[]
    local callback = type(func_name) == 'function' and func_name or api[func_name]

    if callback then
      vim.keymap.set(modes, key_binding, callback, { silent = false, desc = cmds[func_name] and cmds[func_name].desc })
    else
      vim.notify(string.format('No action found for keymap: %s -> %s', key_binding, func_name), vim.log.levels.WARN)
    end
  end
end

```

sudo-tee: Or even create a generic apply keymaps and use it here and in the input_window and output_window  setup

lua/opencode/config.lua
```diff
@@ -163,6 +167,45 @@ M.defaults = {
...
+    -- Translate legacy function names
+    local api_func_name = legacy_function_map[func_name] or func_name
+    normalized[key] = api_func_name
+  end
+  return normalized
+end
+
 --- Setup function to initialize or update the configuration
 --- @param opts OpencodeConfig
 function M.setup(opts)
```
sudo-tee, line 177: With the new format the setup and normalization could become something like this

```lua
--- Convert legacy keymap format to new format
local function normalize_keymap(legacy_config, filter_functions)
  local converted = {}
  for func_name, key in pairs(legacy_config) do
    local api_name = func_name == 'submit' and 'submit_input_prompt' or func_name
    if not filter_functions or vim.tbl_contains(filter_functions, api_name) then
      converted[key] = { api_name }
    end
  end
  return converted
end

--- Get function names from keymap config
local function get_function_names(keymap_config)
  local names = {}
  for _, config in pairs(keymap_config) do
    if type(config) == 'table' and config[1] then
      table.insert(names, config[1])
    end
  end
  return names
end

--- Setup function to initialize or update the configuration
--- @param opts OpencodeConfig
function M.setup(opts)
  opts = opts or {}

  if opts.default_global_keymaps == false then
    values.keymap.editor = {}
  end

  -- Handle legacy keymap format
  if opts.keymap and (opts.keymap.global or opts.keymap.window) then
    vim.notify('opencode.nvim: Legacy keymap format detected. Consider migrating to new format.', vim.log.levels.WARN)

    if opts.keymap.global then
      opts.keymap.editor = normalize_keymap(opts.keymap.global)
    end

    if opts.keymap.window then
      opts.keymap.input_window = normalize_keymap(opts.keymap.window, get_function_names(values.keymap.input_window))
      opts.keymap.output_window = normalize_keymap(opts.keymap.window, get_function_names(values.keymap.output_window))
    end
  end

  -- Merge configurations
  values = vim.tbl_deep_extend('force', values, opts)
end

```

cameronr:
Thanks for all of the comments. Totally agree on getting the format right now so it's worth extra effort to get it there. I also had the same thought about the need to support functions in keymaps (I want my submit keymap, <c-s>, to also exit insert mode). I also wondered if it made sense to split input / output keymaps but wasn't sure... seeing it now, it's definitely much cleaner!

I'll take a look at all of the comments later today.

lua/opencode/ui/session_formatter.lua
```diff
@@ -97,9 +97,9 @@ function M._format_permission_request()
   M.output:add_line(
     string.format(
       '> Accept `%s`    Always `%s`    Deny `%s`',
-      config.keymap[scope].permission_accept,
-      config.keymap[scope].permission_accept_all,
-      config.keymap[scope].permission_deny
+      config.keymap.permission.accept,
```
sudo-tee: One last thing.

The scope there was meant for the keymap to display either the editor like (<leader>opa) if the opencode window is not focused and the permissions one ('a', 'A', 'd') if focused.

cameronr:
@sudo-tee Ok, I think I incorporated all of the feedback. Let me know if there's anything else I should change.

lua/opencode/keymap.lua
cameronr: Went with the generic approach so input/output_window call into keymap

cameronr:
Oh yeah, a few notes:

1. config.lua mentions an api for `open_configuration_file` but it's missing so there's a warning about that keymap

1. `providers` and `custom_commands` are missing from config. I marked them as optional in types but wanted to confirm that's right. Maybe it's better to give them default settings in config?

1. There are also a number of missing fields from `context`. Should those fields be marked as optional or should there be defaults provided?

1. `completion.file_sources` is missing `cache_timeout`

sudo-tee:
1. config.lua mentions an api for `open_configuration_file` but it's missing so there's a warning about that keymap 
 
This is a legacy option you can remove it.

1. `providers` and `custom_commands` are missing from config. I marked them as optional in types but wanted to confirm that's right. Maybe it's better to give them default settings in config?

Also legacy stuff I forgot to remove


1. There are also a number of missing fields from `context`. Should those fields be marked as optional or should there be defaults provided?
 
This one puzzled me. It comes from another PR I merged, they can be removed they are not used.

1. `completion.file_sources` is missing `cache_timeout`

Also comes from a PR I merged without noticing.. It can be removed also

sudo-tee:
> @sudo-tee Ok, I think I incorporated all of the feedback. Let me know if there's anything else I should change.
@cameronr I only found 1 issue which is in the keymap normalization

image

You should filter them by using the function name from the new config. so only the relevant keymaps are applied to the window. Now both window receives all the keymaps.

you can get the function names with a function like this

```lua
--- Get function names from keymap config
local function get_function_names(keymap_config)
  local names = {}
  for _, config in pairs(keymap_config) do
    if type(config) == 'table' and config[1] then
      table.insert(names, config[1])
    end
  end
  return names
end


local function normalize_keymap(legacy_config, filter_functions)
  local converted = {}
  for func_name, key in pairs(legacy_config) do
    local api_name = func_name == 'submit' and 'submit_input_prompt' or func_name
    if not filter_functions or vim.tbl_contains(filter_functions, api_name) then
      converted[key] = { api_name }
    end
  end
  return converted
end
```

and apply them like this 

```lua
 if opts.keymap.window then
      opts.keymap.input_window = normalize_keymap(opts.keymap.window, get_function_names(values.keymap.input_window))
      opts.keymap.output_window = normalize_keymap(opts.keymap.window, get_function_names(values.keymap.output_window))
    end
```

A part from that I'm ready to merge after this tiny fix.

Thanks again :)

sudo-tee:
Amazing work I fixed a small error I was getting when migrating the old format.

I just merged the PR

PR #59, opened 2025-10-09
fix(config): add config metatable

### LINKED ISSUE #58: Config for `input_height` and `window_width` don't work
Just noticed that changing those values now doesn't have any affect even though they do look changed if you check `config.values`.

The issue is that some places, particularly at the top of the file, access the config (and hold onto the result) via:

```lua
local config = require('opencode.config').get()
```

`config.get` just returns `M.values` but `M.values` gets reset during `config.setup` but that places that got the config before setup was called will still have the old table.

The specific issue with `input_height` and `input_width` is caused in `input_window` because it calls `config.get` at the top of the file:

https://github.com/sudo-tee/opencode.nvim/blob/ca0ff9017448db58ee484a687942aab25d503d98/lua/opencode/ui/input_window.lua#L1-L4

and then references that table later:

https://github.com/sudo-tee/opencode.nvim/blob/ca0ff9017448db58ee484a687942aab25d503d98/lua/opencode/ui/input_window.lua#L109-L120

To fix, I'd suggest removing `config.get` and using `setmetatable` so anyone can fetch the config module via `local config = require('opencode.config')` and just call `config.ui.input_height` (or whatever) to always get the correct values.

I'll submit a PR for you to take a look at.

cameronr:
Fixes #58

Makes config access less error prone and more ergonomic


sudo-tee:
I love the solution and it fix a lot of potential issues

PR #60, opened 2025-10-09
fix (base_url): nil guard

DanRioDev:
Small issue for a race condition


DanRioDev:
Sorry, pulled request more than needed

Issue #54, opened 2025-10-09
Incremental output rendering


sudo-tee:

cameronr:
The code is still pretty rough but I was (just this morning) able to use it while working on it, so that's a milestone :)

I'll keep cranking away on it and here are the things on my todo list:
- [x] diff rendering isn't rendering all of the highlights when streaming
- [x] the scrolling gets wonky (at least in my config) sometimes  i think this is fixed now
- [x] first pass on lifecycle management
- [x] i think extmarks are off by one in tool rendering (no space after tool block and next section header)?
- [x] wrong type for _calculate_revert_stats (state.messages is {`info: Message, parts: MessagePart[]}[]|nil`)
- [x] autodiscover json files to test
- [x] fix / commands
- [x] update `next/prev_message` to work with renderer
- [x] hook contextual actions back up
- [x] add appropriate type annotations
- [x] significant code cleanup: a bunch of the code is really gross, particularly how the streaming renderer works with session formatter. once the streaming code is good enough, a bunch of session formatting code can be removed. my plan is to try and just replay all of the events from the session fetch through the streaming renderer.
- [x] message/part caches: it prolly makes sense to have a unified way to track things we have rendered so we know what thing corresponds to which lines.
- [x] the message/part removed code has never been tested (but replacing parts has been)
- [x] testing error handling
- [x] look at undo / redo / snapshots
- [x] highlighting mentions
- [x] robustify network canceling
- [ ] don't render empty assistant message at the end (i think just from stats / tokens update)?
- [x] more complex usage testing
- [x] fix FIXMEs

### Testing 

[`tests/manual/README.md`](https://github.com/cameronr/opencode.nvim/blob/feat/incremental-rendering/tests/manual/README.md#L45-L45) covers how to use the replay system.

When I see something that doesn't look right, I save the streamed events and then check them in the replay tool. Once the issue is fixed, I use `:ReplaySave` to save the desired output and add a unit test

I also have a keymap for reformatting the session using session_formatter:

```
          ['<leader>aR'] = { function() require('opencode.ui.ui').render_output(true) end },
```

As I fix more bugs, there should be less and less need for that.

(moving my [todo list from the pr](https://github.com/sudo-tee/opencode.nvim/pull/62#issuecomment-3394865766) to this issue for tracking)

sudo-tee:
Incremental rendering has been merged

PR #62, opened 2025-10-12
feat(rendering): incremental rendering

sudo-tee:
Introduces an incremental rendering that updates only the changed portion(s) of the buffer instead of re-rendering everything on every change. This reduces CPU usages for large sessions.

This PR also incorporate a functional test suite in `tests/manual` where you can record a run snapshot tests.

### DIFF
--- a/lua/opencode/core.lua
+++ b/lua/opencode/core.lua
@@ -179,17 +179,30 @@ function M.stop()
   if state.windows and state.active_session then
     if state.is_running() then
       M._abort_count = M._abort_count + 1
-      state.api_client:abort_session(state.active_session.id):wait()
+
+      -- if there's a current permission, reject it
+      if state.current_permission then
+        require('opencode.api').permission_deny()
+      end
+
+      local ok, result = pcall(function()
+        return state.api_client:abort_session(state.active_session.id):wait()
+      end)
+
+      if not ok then
+        vim.notify('Abort error: ' .. vim.inspect(result))
+      end
 
       if M._abort_count >= 3 then
         vim.notify('Re-starting Opencode server')
         M._abort_count = 0
         -- close existing server
-        state.opencode_server:shutdown():wait()
+        if state.opencode_server then
+          state.opencode_server:shutdown():wait()
+        end
 
         -- start a new one
         state.opencode_server = nil
-        state.job_count = 0
 
         -- NOTE: start a new server here to make sure we're subscribed
         -- to server events before a user sends a message
@@ -237,7 +250,13 @@ function M.opencode_ok()
   return true
 end
 
+local function on_opencode_server()
+  state.current_permission = nil
+end
+
 function M.setup()
+  state.subscribe('opencode_server', on_opencode_server)
+
   vim.schedule(function()
     M.opencode_ok()
   end)


cameronr:
Comment moved to https://github.com/sudo-tee/opencode.nvim/issues/54#issuecomment-3397782903

sudo-tee:
@cameronr 

First of all great job I had the chance to play a little bit with your PR and it's coming along pretty nicely 


 I have a couple of commits on my branch named the same as yours for the PR but I cannot get them to show on this PR.  There must be something I do wrong, or maybe I need access to the branch on your repo.

You should be able to rebase them without issues as they don't touch the streaming part. 

I will be out for a couples of days so might not have to much time keeping an eye on the progress.

Keep up the good work on this. It's gonna be great.

Thanks again

cameronr:
No worries, I'll pull those changes in.

Glad to hear it's looking good! I'm having fun so I'll keep chipping away at it.

ps: you might want to update the @ as my handle has an r at the end... or maybe cameron wants to join in :)

cameron:
eMacs4lyfe

sudo-tee:
@cameronr 

I've added new commits to render the selected text and also to unify the part rendering.

so there is no need to have 2 seperate methods  for formatting full session or streaming parts :)

You can see them here

https://github.com/cameronr/opencode.nvim/compare/feat/incremental-rendering...sudo-tee%3Aopencode.nvim%3Afeat/incremental-rendering

cameronr:
Oh nice! I had been waiting for the streaming rendering to be good enough to switch over but it seems like we're there.

Thanks for fixing the selection context issue! I didn't have any tests cases for that but I'll add a replay for that.

I also noticed the header separator change. No problem on that change but it does break the unit tests.

I'll add a script to regenerate all of the save replay data (only to be used when making this kind of format change) and I'll update the test data.

I'll also add a github workflow to run the tests on submit/pr.

cameronr:
@sudo-tee I'm going to start cleaning up `session_formatter` (unify the code between full session loading and streaming events more, clean up the M.output hacks used by streaming currently). My quick thinking is:

1. `session_formatter` should be renamed `formatter` and it shouldn't store any state
2. `streaming_renderer` should be renamed `renderer`
3. `output_renderer` should be removed (at some point)
4. the full session loading code should move to `renderer`
5. `renderer` uses `formatter` to format the messages (as it does today, it's just helpful to me to have a clear idea of what each piece is doing)

sudo-tee:
> @sudo-tee I'm going to start cleaning up `session_formatter` (unify the code between full session loading and streaming events more, clean up the M.output hacks used by streaming currently). My quick thinking is:
> 
> 1. `session_formatter` should be renamed `formatter` and it shouldn't store any state
> 2. `streaming_renderer` should be renamed `renderer`
> 3. `output_renderer` should be removed (at some point)
> 4. the full session loading code should move to `renderer`
> 5. `renderer` uses `formatter` to format the messages (as it does today, it's just helpful to me to have a clear idea of what each piece is doing)


This seems pretty reasonable to me 😁

sudo-tee:
@cameronr 

I added a managed event `file.edited` to refresh files when edited.

I also took the liberty to create a dedicated class to do message data lookups. A big chunk of the rendered was just looking up for data.  I created a `MessageMap` with tests class that will do the actual indexing and searching of parts/messages callId.

I also fixed some issues with The Message type, since the migration to server api the types were not correct. I renames Message to OpencodeMessage, because my lsp was mixing it with an internal type of libuv.

I realized (before the refactor) that the contextual actions were not working properly.

Thanks :)

cameronr:
Nice! That's a really nice improvement; cleans things up quite a bit. And love the tests!

I think context actions, undo/redo/snapshots, and prev/next message are the big remaining things that need to be hooked back up. output.lua is related and that's the code I'm looking at now so hoping to make some progress on how it should all fit together.

It's getting there!

cameronr:
@sudo-tee I'm thinking about changing how we process messages / parts from a full session. Right now, there's two different code paths: the full session code path and the event code path. They leverage a lot of the same functionality but they're still subtly different. As one small example, the session reading code wasn't setting `state.last_user_message` anywhere.

It won't be quite as efficient (we'll end up looking up some indexes more often) but i think that the single code path simplicity is worth the extra lookups. Fortunately, MessageMap makes any perf penalty there prolly irrelevant :)

It would involve removing `formatter._format_messages` and replacing it with a loop in `renderer._read_full_session_data`:

```lua
  for i, msg in ipairs(session_data) do
    if state.active_session.revert and state.active_session.revert.messageID == msg.info.id then
      ---@type {messages: number, tool_calls: number, files: table<string, {additions: number, deletions: number}>}
      local revert_stats = M._calculate_revert_stats(state.messages, i, state.active_session.revert)
      local output = require('opencode.ui.output'):new()
      formatter._format_revert_message(output, revert_stats)
      M.write_output(output)

      -- FIXME: how does reverting work? why is it breaking out of the message reading loop?
      break
    end

    -- only pass in the info so, the parts will be processed as part of the loop
    -- TODO: remove part processing code in formatter
    M.on_message_updated({ info = msg.info })

    for j, part in ipairs(msg.parts or {}) do
      M.on_part_updated({ part = part })
    end

    -- FIXME: not sure how this error rendering code works when streaming
    -- if msg.info.error and msg.info.error ~= '' then
    --   vim.notify('calling _format_error')
    --   M._format_error(output, msg.info)
    -- end
  end
```

That way, we're always adding things exactly the same way. It's still possible for streaming to have bugs because of deltas / part replacement but I think the way to have the code be the most robust is to minimize the code paths.

I think it'll also help with cleaning up the types (and maybe even reduce the amount of index shuttling we're having to do) since there will only be one code path.

Does that sound ok?

sudo-tee:
> @sudo-tee I'm thinking about changing how we process messages / parts from a full session. Right now, there's two different code paths: the full session code path and the event code path. The leverage a lot of the same functionality but they're still subtly different. It won't be quite as efficient (we'll end up looking up some indexes more often) but i think that the single code path simplicity is worth the (likely tiny perf penalty).
> 
> It would involve removing `formatter._format_messages` and replacing it with a loop in `renderer._read_full_session_data`:
> 
> ```lua
>   for i, msg in ipairs(session_data) do
>     if state.active_session.revert and state.active_session.revert.messageID == msg.info.id then
>       ---@type {messages: number, tool_calls: number, files: table<string, {additions: number, deletions: number}>}
>       local revert_stats = M._calculate_revert_stats(state.messages, i, state.active_session.revert)
>       local output = require('opencode.ui.output'):new()
>       formatter._format_revert_message(output, revert_stats)
>       M.write_output(output)
> 
>       -- FIXME: how does reverting work? why is it breaking out of the message reading loop?
>       break
>     end
> 
>     -- only pass in the info so, the parts will be processed as part of the loop
>     -- TODO: remove part processing code in formatter
>     M.on_message_updated({ info = msg.info })
> 
>     for j, part in ipairs(msg.parts or {}) do
>       M.on_part_updated({ part = part })
>     end
> 
>     -- FIXME: not sure how this error rendering code works when streaming
>     -- if msg.info.error and msg.info.error ~= '' then
>     --   vim.notify('calling _format_error')
>     --   M._format_error(output, msg.info)
>     -- end
>   end
> ```
> 
> That way, we're always adding things exactly the same way. It's still possible for streaming to have bugs because of deltas / part replacement but I think the way to have the code be the most robust is to minimize the code paths.
> 
> Does that sound ok?

Yes that sounds reasonable to me. It's always good to have a common code path 

I could explain a little more how revert works. But I only have access to my phone for the next couple of days.

But essentially opencode will store the message Id of where the revert happens.  And the rest of the messages don't technically exists. So the stats are calculated from that revert point.  If you call /revert another time it will move up to the previous messageId.  and so one.

At the moment this is buggy in opencode as once the server is closed.. or the opencode tui is closed you cannot revert anymore it gives an error and Delete the files entirely 😵. There are a couple of open bugs on the open code repo about it.


Again I am very grateful for your help on complex topics. Keep up the good work.

cameronr:
Ok, great, I'll keep moving in that direction.

And thanks for the explanation of revert. I'll see if I can figure out how to handle that for session load and when streaming events.

cameronr:
I did a fairly big refactor and introduced a new `RenderState` object that subsumed _part_cache,  _message_map, and _actions to better handle finding parts/messages and tracking their rendered position. I think I'm happy with how that code is coming along :)

Separately, I haven't had a chance to look at snapshots/undo/redo yet but I've noticed a few times where canceling a request didn't result in `job_count == 0`. I have some ideas on that so I'm going to take a look at that next.

sudo-tee:
> I did a fairly big refactor and introduced a new `RenderState` object that subsumed _part_cache,  _message_map, and _actions to better handle finding parts/messages and tracking their rendered position. I think I'm happy with how that code is coming along :)
> 
> Separately, I haven't had a chance to look at snapshots/undo/redo yet but I've noticed a few times where canceling a request didn't result in `job_count == 0`. I have some ideas on that so I'm going to take a look at that next.


Awesome. I think the RenderState reflect the idea I had in my head with The Output class and the MessageMap.  I think you found the pattern I was searching for. Great idea.  For the undo redo I will be back home this weekend and will have a look if you want. 


There is also a couple of bugs that are piling up in the repo that I need to address.


Thanks again

cameronr:
I'll actually be away for much of this weekend so it would be amazing if you were able to look into undo / redo / snapshots!

I'm hoping the big changes are done and now it's mostly about hooking things back up and doing a bunch of testing to build confidence that it's all working.

sudo-tee:
@cameronr 

I'm almost there for the revert part. I only need to make it work with the test replay.

cameronr:
that's great! let me know if you have any trouble with getting the replay testing working (or with the rendering).

one note is that each replay tests both incremental rendering (streamed events) and full session loads (to make sure they render the same as incremental). the full session loads are reconstructed from the events, tho, and that reconstruction doesn't handle some things (like permissions) so some tests need to disable full session testing.

sudo-tee:
@cameronr 

I pushed my fix for undo/redo.

I fixed redo to match the opencode tui behavior

I also added some tests for undo redo

I will have a look at snapshots next.

sudo-tee:
@cameronr 

Snapshots are a custom thing so there is no event for it...  With the proper undo redo behavior. I am wondering if simply calling undo on the message instead of having 2 different workflow with slightly different meaning.. 

Although the snapshot handling is more granular than the undo redo..

cameronr:
Nice! I think we're getting pretty close!

There's a FIXME in output_window about `restore_points`. Can you take a look at that whenever you next have a chance?

cameronr:
I have a todo item for mentions but I'm not sure how they work so could use some help.

In the old code, `output_renderer` called `highlight_all_mentions` after rendering:

https://github.com/sudo-tee/opencode.nvim/blob/fed694100d165f15aceab50a91d874d45274005f/lua/opencode/ui/output_renderer.lua#L155-L160

My reading of the `highlight_all_mentions` code says it would apply the `OpencodeMention` extmark to anything in that output that was like `@word` (where `word` can include `_-./`. That doesn't seem right as many things might match that (lua comments like `@param`, as just one example).

One option might be to only apply the highlight to the text property of a part but I don't have any examples of what I think highlights is attempting to match. Do you have any examples (or have a good way to generate them)?

thanks!

sudo-tee:
> Nice! I think we're getting pretty close!
> 
> There's a FIXME in output_window about `restore_points`. Can you take a look at that whenever you next have a chance?

@cameronr 

I was able to plug restore_points back. But unfortunately I cold not get it to work properly in the replay renderer as it reads the filesystem for data. 

This is where I question if I should keep the feature or not. It's useful but it differs From the Opencode way of doing it.

sudo-tee:
> I have a todo item for mentions but I'm not sure how they work so could use some help.
> 
> In the old code, `output_renderer` called `highlight_all_mentions` after rendering:
> 
> https://github.com/sudo-tee/opencode.nvim/blob/fed694100d165f15aceab50a91d874d45274005f/lua/opencode/ui/output_renderer.lua#L155-L160
> 
> My reading of the `highlight_all_mentions` code says it would apply the `OpencodeMention` extmark to anything in that output that was like `@word` (where `word` can include `_-./`. That doesn't seem right as many things might match that (lua comments like `@param`, as just one example).
> 
> One option might be to only apply the highlight to the text property of a part but I don't have any examples of what I think highlights is attempting to match. Do you have any examples (or have a good way to generate them)?
> 
> thanks!

@cameronr 

The `highlight_all_mentions` is indeed finding `@{subagent}` or the `@path/to/file.lua` and add highlight to them. 

When you type @ in the input_window and select either a file or a subagent , it will call `highlight_all_mentions` also on the input_window to highlight them.

You are right that in the output_window it might be error prone. 

However if you look at the message you at the messages generated you will see them with a range

```json
{
      "messageID": "msg_a06c4164a0015W5xZmeTCOAuhE",
      "source": {
        "text": {
          "value": "@test.txt",
          "end": 8,
          "start": 0
        },
        "path": "/home/francis/Projects/_nvim/opencode.nvim/test.txt",
        "type": "file"
      },
      "mime": "text/plain",
      "type": "file",
      "id": "prt_a06c416520063e9hNFYlVzbCvm",
      "url": "file:///home/francis/Projects/_nvim/opencode.nvim/test.txt",
      "filename": "test.txt",
      "sessionID": "ses_5f9558f7affeMXh93WS3cQpzjB"
    },
```

```json
    {
      "messageID": "msg_a06c60f89001za14H2cqggp6uq",
      "type": "agent",
      "id": "prt_a06c60f89003AyAUE4WJWAXoHK",
      "source": {
        "value": "@foo",
        "end": 4,
        "start": 0
      },
      "name": "foo",
      "sessionID": "ses_5f9558f7affeMXh93WS3cQpzjB"
    },
```

The source part indicates the range in the text (this is how they do in internally in opencode)

Maybe we can use theses to generate the proper highlight ?  or maybe use the value part for the match something.

UPDATE:

It's more complicated than it looks as the user prompt is rendered before the other message parts.

sudo-tee:
@cameronr 

I pushed a basic implementation of mention highlight for only user prompt, however it still can highlight "wrong" mentions.

The output also need a test which I had not time to work on. I wanted to have something that would help you getting unstuck

cameronr:
Oh, great. I'll take a look at the code you added. 

In my own sessions, I see examples of mentions in my own messages so I can create the test for that.

Is that the only place mentions need to be highlighted or can they also show up / need to be highlighted in the assistant responses?

sudo-tee:
As far as I know it's only for user messages

cameronr:
I have the code to render file mentions using the ranges in a part (by rerendering the earlier user prompt part) but I think there's bug in opencode where it sends back a range starting at 0 sometimes. You can see it in the `diff.json` test:

https://github.com/sudo-tee/opencode.nvim/blob/18374c114e15703d5e33a18500a9494ade258611/tests/data/diff.json#L97-L99

That causes the highlight to be added to the beginning of the user prompt. Interestingly, the opencode client seems to render that data the same way. Here's a similar session rendered in both:

in opencode.nvim:

Screenshot 2025-10-21 at 17 16 53

in opencode tui:

Screenshot 2025-10-21 at 17 16 47

I can check and make sure the range starts with and `@` so at least it's not highlighting incorrectly but then there will be highlights missing. I guess, in that case, I could fall back to searching the text `value` and highlighting... I'll try that.

As just an aside, I also noticed that the opencode tui client has an off by one when rendering the highlight (I had the same bug initially):
Screenshot 2025-10-21 at 16 14 32

sudo-tee:
This one was a bug in  the plugin itself.. there was a off by one error..  that I just fixed today in one of my commits.. you probably have an outdated file with the error

cameronr:
The off by one is in the opencode tui (not opencode.nvim). I was just pointing out that the opencode tui code doesn't seem very robust in this area.

All of that said, the off by one isn't important (after all, it's not even in our code).

The real issue is that the file ranges are sometimes / often zero which cause erroneous highlighting, both in the code I've just written and in the opencode tui app.

Here's some test data (with only the relevant parts):

```json
  {
    "type": "message.part.updated",
    "properties": {
      "part": {
        "sessionID": "ses_5f6d8a858ffeahzN9rkzzAlFMu",
        "messageID": "msg_a0929114f001p91aGAr5mfFVfh",
        "id": "prt_a0929114f002IHnVoMagRTlmq2",
        "type": "text",
        "text": "this is a mention test @diff-test.txt and @a-empty.txt. you can just say 1"
      }
    }
  },
  {
    "type": "message.part.updated",
    "properties": {
      "part": {
        "url": "file:///Users/cam/tmp/a/diff-test.txt",
        "sessionID": "ses_5f6d8a858ffeahzN9rkzzAlFMu",
        "mime": "text/plain",
        "messageID": "msg_a0929114f001p91aGAr5mfFVfh",
        "id": "prt_a09291150006rGf7yI937pvN4l",
        "type": "file",
        "filename": "diff-test.txt",
        "source": {
          "path": "/Users/cam/tmp/a/diff-test.txt",
          "type": "file",
          "text": {
            "start": 0,
            "end": 14,
            "value": "@diff-test.txt"
          }
        }
      }
    }
  },
  {
    "type": "message.part.updated",
    "properties": {
      "part": {
        "url": "file:///Users/cam/tmp/a/a-empty.txt",
        "sessionID": "ses_5f6d8a858ffeahzN9rkzzAlFMu",
        "mime": "text/plain",
        "messageID": "msg_a0929114f001p91aGAr5mfFVfh",
        "id": "prt_a092911500032HS2DGrZwukRwQ",
        "type": "file",
        "filename": "a-empty.txt",
        "source": {
          "path": "/Users/cam/tmp/a/a-empty.txt",
          "type": "file",
          "text": {
            "start": 0,
            "end": 12,
            "value": "@a-empty.txt"
          }
        }
      }
    }
  },
```

in opencode.nvim:

Screenshot 2025-10-21 at 17 16 53

and in the opencode tui app:

Screenshot 2025-10-21 at 17 16 47

and here's what it looks like after I implemented the fallback I mentioned above (if `source.text.start == 0` and there's not an `@` at that position in the text, then search for `value`):

Screenshot 2025-10-21 at 18 18 37

cameronr:
Wanted to check on a few things:

1. I just pushed my mention highlighting change. Let me know if it doesn't work the way you want/expect
2. I updated extmarks in the formatter (and elsewhere) to be 0 indexed instead of 1 indexed to be consistent with the nvim api. It shouldn't have an impact but I just wanted to give you a heads up.
3. I'm wondering if we should make the `on_data_rendered` callback that's used for markdown rendering have a configurable debounce delay in case people want to tune it (e.g. markview seems to work better with a longer debounce). Maybe something like `on_data_rendered_debounce`?
4. Should the types `OpencodeMessagePart` and `MessagePart` be merged into just `OpencodeMessagePart`?

sudo-tee:
> Wanted to check on a few things:
> 
> 1. I just pushed my mention highlighting change. Let me know if it doesn't work the way you want/expect

It seems to work fine however the agent mentions were missing so I added them


> 2. I updated extmarks in the formatter (and elsewhere) to be 0 indexed instead of 1 indexed to be consistent with the nvim api. It shouldn't have an impact but I just wanted to give you a heads up.

Thsnkd it should clears thing up. I was always confused by their indexes.

> 3. I'm wondering if we should make the `on_data_rendered` callback that's used for markdown rendering have a configurable debounce delay in case people want to tune it (e.g. markview seems to work better with a longer debounce). Maybe something like `on_data_rendered_debounce`?

This is a good idea, I added a config option in my branch for it 
image


> 4. Should the types `OpencodeMessagePart` and `MessagePart` be merged into just `OpencodeMessagePart`?

I also noticed that there were duplication in the types yesterday. Merged the 2 types in `OpencodeMessagePart` in my branch

cameronr:
Those are all great changes! I'll move the `on_data_rendered` to nest under `output.rendering`

cameronr:
@sudo-tee Just checked in my change that will restart the opencode server if the user cancels three times in a row. Let me know if it looks ok or if there are any changes that I should make.

There isn't much left on my todo list for incremental rendering. I'll keep testing it but I'm wondering if there's other stuff on your list or if we should start thinking about how to get it released?

sudo-tee:
@cameronr 

The fix looks great to me :). 

I don't have anything that I want to add at the moment.

The sooner we can merge this the better. 

As for the release. I will try to use it as much as possible in the next day or 2. 

There is also some conflicts with master, but I don't think they is any major issues.

Thanks again for the big effort :)

cameronr:
@sudo-tee Oh yeah, I'm curious why the initial setup call is wrapped in `vim.schedule`:

https://github.com/sudo-tee/opencode.nvim/blob/fb420c8c4d58dce58856ee8fc8d5aeb9c2ccd6e4/lua/opencode/init.lua#L3-L18

It obviously works but it does make it lazyloading quite a bit trickier since the api commands aren't guaranteed to exist after the setup call (they're added whenever the event loop next gets to it). It doesn't look like there's anything in heavyweight done in any of those calls so I think it would be ok to remove the vim.schedule. What do you think?

sudo-tee:
> @sudo-tee Oh yeah, I'm curious why the initial setup call is wrapped in `vim.schedule`:
> 
> https://github.com/sudo-tee/opencode.nvim/blob/fb420c8c4d58dce58856ee8fc8d5aeb9c2ccd6e4/lua/opencode/init.lua#L3-L18
> 
> It obviously works but it does make it lazyloading quite a bit trickier since the api commands aren't guaranteed to exist after the setup call (they're added whenever the event loop next gets to it). It doesn't look like there's anything in heavyweight done in any of those calls so I think it would be ok to remove the vim.schedule. What do you think?

There use to be blocking stuff in there, like fetching the config on load but not anymore.

I think it's safe to remove

sudo-tee:
@cameronr 

I fixed a bug in the permissions prompt. When the focus changes to the editor it should display the global shortcuts and not the buffer local ones. I also added the keymaps for permission on the output buffer as well.

I also fixed the debug helper debug_message to look for the nearest message upward instead of a fixed line.  It's way more useful this way

cameronr:
@sudo-tee I have the code to fix the last functional todo on my list (not rendering empty assistant or user messages):

https://github.com/sudo-tee/opencode.nvim/compare/feat/incremental-rendering...cameronr:opencode.nvim:feat/deferred-msg-render

So, instead of this:

Screenshot 2025-10-23 at 07 54 54

It renders this:

Screenshot 2025-10-23 at 07 55 00

##

It requires deferring the message header rendering until we have a part (or error) that needs it. The tests pass (except for where there are correct changes) and the code feels clean enough but there's some risk that there's some sequence of messages / parts that I haven't seen/accounted for. I'm trying to decide it's better to hold off on this code for now.

How important is it to not render those occasional empty user/assistant message headers?

I think what probably makes the most sense is to wait on this code and get the current version out (after some more testing and no more changes) and make sure that's working well before adding this.

What do you think?

sudo-tee:
Fair point, 

It's better to release it, and add things later on. 

I am doing a final review of the code, since it's a massive change I want to get comfortable with it. But it looks good and stable for me in my tests

lua/opencode/ui/topbar.lua
```diff
@@ -86,20 +86,35 @@ local function get_session_desc()
...
   end)
 end
 
+local function on_change(_, _, _)
+  M.render()
+end
+
+function M.setup()
+  state.subscribe('current_mode', on_change)
+  state.subscribe('active_session', on_change)
```
sudo-tee, line 113: maybe we should add current_model as well.

in case it changes ?

lua/opencode/server_job.lua
```diff
@@ -17,6 +18,18 @@ local function handle_api_response(response, cb)
   end
 end
 
+function M.get_unresolved_requests()
```
sudo-tee: I don't think it's used anymode

lua/opencode/server_job.lua
```diff
@@ -91,23 +136,35 @@ function M.stream_api(url, method, body, on_chunk)
   return curl.request(opts)
 end
 
+---Forcibly reject any pending requests (they sometimes get stuck
+---after an api abort)
+function M.cancel_all_requests()
```
sudo-tee: Same goes for this method, it does not seem to be called anywhere

sudo-tee:
@cameronr 

I did a rebase from main on my branch. It should clear the conflicts warning

sudo-tee:
I'm gonna continue my tests tomorrow. If I don't see any breaking things. I will merge. 

It will feel good to have this merged 

I might need to squash it as one big feature so it's easier to rollback if anything happens. It will also be better for the main branch history

cameronr:
That's very exciting! I'll keep testing as well and let you know if I see any issues. It will be great to have this in main!

cameronr:
I fixed a bug in config (the push right above this comment) where default keymaps were overwriting user keymaps if `keymap_prefix` was specified and the user set their keymap using the new prefix, e.g.:

```lua
      keymap_prefix = '<leader>a',
      keymap = {
        editor = {
          ['<leader>ai'] = { function() require('opencode.core').open({ new_session = false, focus = 'input' }) end, mode = { 'n', 'v' } }, -- open input without insert mode
```

It was overwriting my `<leader>ai` keymap with the default one

cameronr:
Made another fix to strip out ansi escape sequences from bash/code blocks:

so this:
```
▌^[[32mSuccess: ^[[0m    33
▌^[[31mFailed : ^[[0m    2
▌========================================
▌Tests Failed. Exit: 1
▌^[[0;31m✗ Unit tests failed^[[0m
▌------------------------------------------------
▌
▌^[[0;31m======== TEST FAILURES SUMMARY ========^[[0m
▌^[[0;31mFound 2 failing test(s):^[[0m
▌
▌^[[0;31mFAILED TEST:^[[0m       renderer unrendered messages
▌_remove_message_from_buffer skips unrendered message
▌^[[0;31mFAILED TEST:^[[0m       renderer unrendered messages
▌_rerender_part skips if parent message is unrendered
```

becomes:
```
▌Success:        33
▌Failed :        2
▌========================================
▌Tests Failed. Exit: 1
▌✗ Unit tests failed
▌------------------------------------------------
▌
▌======== TEST FAILURES SUMMARY ========
▌Found 2 failing test(s):
▌
▌FAILED TEST:    renderer unrendered messages
▌_remove_message_from_buffer skips unrendered message
▌FAILED TEST:    renderer unrendered messages _rerender_part skips if
▌parent message is unrendered
```

sudo-tee:
@cameronr 

I was testing the PR for a while this morning and found 2 things.

When there is a lot of events the UI locks up. I ran some profiling and saw this

image

I think nvim is choking to render this many events. We might need a "scheduler" , I tried to put more vim.schedule bit it did not help.

You can use a prompt like this `I just want to test the markdown rendering, please write as long session with codeblocks
` with a model like gpt 4.1


I also saw a lot of markdown flashing with render-markdown. I think we might need to put a larger debounce for markdown rendering. 

I will play a bit more with it and try to address the issues I found.

sudo-tee:
I did push  some perf improvements on the Output:add_line(s) which helps considering how often the add_line is called.

I also used vim.schedule_wrap in every callback in the event_manager.  

Now the real bottleneck shows and it's the set_lines which makes sense, this causes nvim ui to lock.

cameronr:
Oh, interesting. Whenever there's an issue (perf or otherwise), you can use `debug_helpers.save_captured_events` to save all of the events to a json file that can be loaded into the replay tool. That can be really helpful to test performance improvements with the same data over and over again.

I'll also do as you suggested and generate some larger sessions.

edit: I don't see any [changes](https://github.com/sudo-tee/opencode.nvim/compare/feat/incremental-rendering...cameronr:opencode.nvim:feat/incremental-rendering). Are you sure your changes got pushed up or are they in a different branch?

sudo-tee:
This is weird, 

https://github.com/sudo-tee/opencode.nvim/commits/feat/incremental-rendering/

I see 2 commits there...

cameronr:
Huh, i guess just some GH weirdness. I pulled the latest changes :)

sudo-tee:
I did a comparison to main and there you can really see the issue on the incremental rendering branch

Incremental rendering
https://github.com/user-attachments/assets/9a246fec-f234-4963-a061-f7660b4624ad

Main
https://github.com/user-attachments/assets/27f1d5d4-674f-413c-b310-1bb501561be6

You can see that the ui locks up when rendering..

cameronr:
I have been testing with this [very large session file](https://github.com/user-attachments/files/23130246/huge-session.gz) and using the session replay with `ReplayFullSession` and I think I've found the issue.

First some background:

1. The snacks profiling data, while very helpful, sometimes hides the true cause performance loss. I've found it most helpful as a pointer in the right direction but then I have to actually try changing/removing things to confirm the specific issue.
2. Adding this to snacks.profiler config helps as it includes vim functions in the profiler view: `globals = { 'vim' }`
3. Adding `vim.schedule` can help in some specific cases but it's a bit of a bandaid for performance issues because it doesn't reduce the work done on the UI thread.
4. To profile with my full config, I add this to my opencode.nvim config (only works in the opencode.nvim directory):
```lua
    vim.api.nvim_create_user_command('OpencodeReplay', function()
      vim.opt.runtimepath:append('.')
      require('tests.manual.renderer_replay').start({ set_statuscolumn = false })
    end, {}),
```

With all of that background, this is my baseline profiling result:

Screenshot 2025-10-24 at 11 23 38

The time spent in `deepcopy` really jumps out and I traced it down to the `deepcopy` in state.append:

https://github.com/sudo-tee/opencode.nvim/blob/e93708de4e1e25fa326193a940cb1f0a45dcd257/lua/opencode/state.lua#L138-L152

Not doing a deepcopy on old gives dramatically faster perf:
Screenshot 2025-10-24 at 11 26 14

I don't think the deepcopy is necessary (all test pass without it) but I'm going to take a closer look and do a quick pass on state generally. I should have a fix committed in a bit.

cameronr:
I looked more close at `state.messages` and `state.append`. No one's actually subscribed to `state.messages` so all of the deepcopy work is throwaway work anyway.

I think the best fix is to not use `state.append` when adding messages. It does mean that the state notification system won't work for new messages, I think that's fine because:

1. EventManager already has new message notifications (but server only)
2. renderer is who cares the most about messages and it's already handling things internally, including notifying other modules if they need to update based on messages
3. in the future, we could add a separate new_message notification if need. in fact, `state.last_message` is kind of that already

I also don't think we need the `vim.schedule` in `EventManager:emit`.

We should also inspect other uses of `state.append/remove` to make sure they aren't sources of extra / not useful work.

sudo-tee:
Thanks for the details, I didn't know about the profiler options.

Here is the profile.  The Issue I see is really calling the set_lines

image


As per the deepcopy,  for the old value, it's there because if you give a reference you will have the new value.

I see that this could be a problem, but in my profiling this is not really a hot path.

The hot path is the part rerendering as the part grows, overwrites the lines of the same part over and over again. I think we could try to only append new lines instead.

cameronr:
Hmm, I think that means there are several performance issues.

I think the `deepcopy` is definitely a problem because it copies the entire messages table, along with every part, every time a message is added. That will get slower and slower the more messages you have. My `huge_session` file has 1021 messages so as more and more messages are added, it gets quite slow.

Your profile results don't really make sense to me as `nvim_buf_set_lines` should be fast so there must be something I'm not understanding / something else going on.

To help debug, can you do the following:

1. Can you share the captured events from a session that shows this? It'll be much easier if we're working with the same data. If it's a session that already exists, you can temporarily modify `renderer._render_full_session_data` to add messages and parts to `event_manager.captured_events`:
```lua
function M._render_full_session_data(session_data)
  M.reset()

  -- disable auto-scroll, makes loading a full session much faster
  M._disable_auto_scroll = true

  local revert_index = nil

  local event_manager = state.event_manager

  for i, msg in ipairs(session_data) do
    -- output:add_lines(M.separator)
    -- state.current_message = msg

    if state.active_session.revert and state.active_session.revert.messageID == msg.info.id then
      revert_index = i
    end

    M.on_message_updated({ info = msg.info }, revert_index)

    table.insert(event_manager.captured_events, { type = 'message.updated', properties = { info = msg.info } })

    for _, part in ipairs(msg.parts or {}) do
      table.insert(event_manager.captured_events, { type = 'message.part.updated', properties = { part = part } })

      M.on_part_updated({ part = part }, revert_index)
    end
  end

  if revert_index then
    M._write_formatted_data(formatter._format_revert_message(state.messages, revert_index))
  end

  -- re-enable Auto-scroll
  M._disable_auto_scroll = false
  M._scroll_to_bottom()
end
```

and then you can save the events with `require('opencode.ui.debug_helper').save_captured_events('some-file.json')`

2. What happens if you comment out `vim.api.nvim_buf_set_lines(windows.output_buf, start_line, end_line, false, lines)
` from `output_window`? I know it won't update the buffer, but I'm curious if that changes the performance or nt.
3. With the replay file, do you see the same performance issue when using the replay tool?

sudo-tee:
If I comment the set_lines the ui does not lock up and the thinking animation is smooth


[perf.json](https://github.com/user-attachments/files/23133835/perf.json)

I included a perf file. 

When I run the file in the replay renderer everything looks smooth.


I will try to load the ReplayRender with my full config

cameronr:
Very helpful! This is what I see when I replay `perf.json` with my full config (not the stripped down replay init):

Screenshot 2025-10-24 at 12 38 39

So could be related to config? Or could still could be something different the replay runner is doing vs the main code.

sudo-tee:
With my full config I have the issue now also in replay.
With the same profile as you

image

sudo-tee:
In the stand alone replay,  there is no syntax highlighting, no markdown rendreing, no lsp etc. So my guess is that somewhere, not seen in the profile is being upset about having set_lines being spammed over and over again.

cameronr:
Hmm, I suspect some interaction with your config as I have markdown rendering, syntax highlighting, etc with the same file without any performance issues:

Screenshot 2025-10-24 at 12 58 29

Screenshot 2025-10-24 at 12 58 33

Is your nvim config shared anywhere?

sudo-tee:
If I call `ReplayAll 16` with 16ms intrervals let's say it becomes way smoother.  
Which leads me to think we need might our own `buffering/scheduling` 

We accumulate the buffer writes, and with a timer we write the content one changes ate the time with a small delay

sudo-tee:
Yes you can see it here

https://github.com/sudo-tee/dots/tree/main/apps/nvim/config


It might also be my terminal emulator.

I'm using Wezterm and my dev-env is WSL (I can only use windows at work unfortunately)

EDIT: 

I tried another terminal and it's the same.

cameronr:
Oh, I think I might have been confused.

I was testing raw rendering speed via `:ReplayFullSession` and `:ReplayAll 0`. Let's first make sure those aren't slow for you.

If those are slow, then the first place to look is in our own code (like the state.append thing). If those are fast, then what happens if you disable render-markdown entirely?

I'm also going to try out your config and see if I notice anything.

sudo-tee:
`:ReplayFullSession` and `ReplayAll 0` are fast for me.

But they don't reflect "real life" usage when opencode spams event.

sudo-tee:
If I play with the markdown debouce and put a big value like 1250ms it's a bit better

cameronr:
Right, I just wanted to make sure they were fast as a baseline (If they were slow, then we should start with them).

The next step is to figure out if it's a side effect (like markdown rendering). That would be my first guess.

edit: Ah, sounds like that might be it.

cameronr:
Markdown rendering likely to be heavily dependent on system performance and I agree that if the markdown debounce delay is too short, it can be pretty jarring.

FWIW, with your config, `ReplayAll 5` on my machine  (apple m4) is smooth (although the markdown background does flicker a bit). This test file is probably also particularly sensitive to markdown performance impacts because it's all markdown.

If you disable render-markdown (but leave treesitter markdown on), what's the experience like?

So maybe we should just make the default debounce larger?

Lastly and not important at all, but if you wanted, you could add a dev block to your lazy.nvim config that specifies the directory where you do your local plugin development, e.g.:

```lua
  performance = {
  --- ...
  },
  dev = {
    path = '~/Dev/neovim-dev/',
    fallback = false,
  },
```

Then you can just set `dev = true,` in a plugin spec, and it'll use the files in that dir.

cameronr:
I'm going to try implementing a text delta update when replacing a part. I had tried it early on but it introduced a bunch of bugs and I wasn't sure a dedicated code path was worth it.

I looked at perf.json, tho, and and a single part occurs over 2000:

```
❯ rg prt_ perf2.json | sort | uniq -c | sort
   1         "id": "prt_a17b4dc4c002cnV3qMuzC78qN0",
   1         "id": "prt_a17b4dc4f001l4IPCU1KYr2agj",
   1         "id": "prt_a17b4dc4f002OaKYJovbdWlZKb",
   1         "id": "prt_a17b4dc50001ZN7SRfxFasNXmK",
   1         "id": "prt_a17b4f1620013BhJdS3rKb37ML",
   1         "id": "prt_a17b51fb5001OJbu79hmP1ZzOv",
2189         "id": "prt_a17b4f16b001dtQr5PNp0YCP8u",
```

So I think it's worth another look.

sudo-tee:
This was also my conclusion.  The larger the part is growing the worse it gets.

cameronr:
Ok, that wasn't too bad actually. All of that work we did cleaning the code up (and all of the unit tests!) made that pretty straight forward.

So now, when replacing a part, the code only sets the changed lines. I did some quick stats:

before the change, when running `perf.json` `output_window` tracked changing 316k lines:
```
Event 2204/2204: session.updated ses_5e84ca285ffeY3zxSfonM8rH4T lines_set: 316076
```

And after the change:
```
Event 2204/2204: session.updated ses_5e84ca285ffeY3zxSfonM8rH4T lines_set: 38372
```

I think we could potentially go farther (only set the part of the line that changed). I'll look at that but that might get messy, not sure yet.

Want to try out the change and see if you notice any difference?

cameronr:
> This was also my conclusion. The larger the part is growing the worse it gets.

It's funny, we found two totally different performance issues: I found one with too many messages and you found one with a single part that was updated too many times :)

sudo-tee:
Unfortunately I still see some locking. I think it's more the amount of writes rather the the size. It's like the editor needs to breathe. Maybe we can batch the same part

cameronr:
Hmmm...

Is it the same amount of locking up or is it reduced? If it's reduced, any sense on how much it's reduced by?

I just noticed the delta code is still replacing more lines that it should so I'm going to look at that first. But will start thinking about part caching in the back of my mind.

You don't have any autocmds or plugins that trigger on `TextChanged` do you?

cameronr:
Ok, i fixed a bug where it was sometimes resetting the whole part so now it's both setting dramatically fewer lines and even fewer slightly fewer calls.

Original:
```
Event 2204/2204: session.updated ses_5e84ca285ffeY3zxSfonM8rH4T lines_set: 316076 set_calls: 2193
```

Latest
```
Event 2204/2204: session.updated ses_5e84ca285ffeY3zxSfonM8rH4T lines_set: 3868 set_calls: 1933
```

Can you try the following and see if any of them make a difference

1. Just this latest version. I've also included the debugging print code in output_window and the replay runner.
2. What happens if you turn off autocmds during wrapping `nvim_buf_set_lines` in `output_windows.set_lines` with `eventignore`:
```lua
  local old_eventignore = vim.o.eventignore
  vim.o.eventignore = 'all'
  vim.api.nvim_set_option_value('modifiable', true, { buf = windows.output_buf })
  -- vim.notify(vim.inspect(lines))
  vim.api.nvim_buf_set_lines(windows.output_buf, start_line, end_line, false, lines)
  vim.api.nvim_set_option_value('modifiable', false, { buf = windows.output_buf })
  vim.o.eventignore = old_eventignore
```
3. Is there any difference if you comment out setting extmarks and make output buf modifiable so you don't have toggle modifiable on and then off each time?
4. Is there any difference if you disable rendering of the footer, topbar, and loading animation?

I'm trying to figure out what's different between loading the full session (which makes all of the same event calls and then all of the same buffer calls) and having the events coming in more slowly.

It may be that we need to batch up changes but I'd really like to know which limit we're hitting to make sure we fix it.

cameronr:
I spent more time looking at this and I think it's related to treesitter. When able, can you try just commenting out the treesitter register line:

https://github.com/sudo-tee/opencode.nvim/blob/e93708de4e1e25fa326193a940cb1f0a45dcd257/lua/opencode/ui/ui.lua#L86

If that works, can you comment that line back in but try turning off treesitter at the start of `renderer.on_part_updated` via `vim.treesitter.stop(state.windows.output_buf)` and then calling `vim.treesitter.start(state.windows.output_buf)` in trigger_on_data_rendered? That will debounce the async treesitter processing. The experience isn't great because the colors disappear but I just want to see if that actually makes the perf issue go away.

I kind of wonder if the faster we make the event processing, the worse it is because that just queues up more work for treesitter and eventually we hit some limit. That would also make sense why doing `ReplayFullSession` or `ReplayAll 0` doesn't exhibit the problem because we do all of our work in a tight loop (without any vim.schedules/defer_fn) so the main loop doesn't have a chance to get treesitter looking at each of our tiny changes.

Assuming that's it, I don't think it's a good experience to turn off treesitter while we're waiting for more updates. And since I don't know of another way to control it, I think we'll have to do some kind of throttling. I have a first pass at the code but I'm still figuring out a good way to test it. Should have something tomorrow.

sudo-tee:
> Ok, i fixed a bug where it was sometimes resetting the whole part so now it's both setting dramatically fewer lines and even fewer slightly fewer calls.

It is already a lot better. there is no more ui lockup for multiple seconds. The rendering still feels a but heavy  and nvim is hitting 100%. But at least the text is written in the buffer.

> 
> Original:
> 
> ```
> Event 2204/2204: session.updated ses_5e84ca285ffeY3zxSfonM8rH4T lines_set: 316076 set_calls: 2193
> ```
> 
> Latest
> 
> ```
> Event 2204/2204: session.updated ses_5e84ca285ffeY3zxSfonM8rH4T lines_set: 3868 set_calls: 1933
> ```
> 
> Can you try the following and see if any of them make a difference
> 
> 1. Just this latest version. I've also included the debugging print code in output_window and the replay runner.
> 2. What happens if you turn off autocmds during wrapping `nvim_buf_set_lines` in `output_windows.set_lines` with `eventignore`:
> 
> ```lua
>   local old_eventignore = vim.o.eventignore
>   vim.o.eventignore = 'all'
>   vim.api.nvim_set_option_value('modifiable', true, { buf = windows.output_buf })
>   -- vim.notify(vim.inspect(lines))
>   vim.api.nvim_buf_set_lines(windows.output_buf, start_line, end_line, false, lines)
>   vim.api.nvim_set_option_value('modifiable', false, { buf = windows.output_buf })
>   vim.o.eventignore = old_eventignore
> ```


> 3. Is there any difference if you comment out setting extmarks and make output buf modifiable so you don't have toggle modifiable on and then off each time?
> 4. Is there any difference if you disable rendering of the footer, topbar, and loading animation?

For the other points, there was no perceivable difference when removing extmarks, footer, loading animation or topbar.


> I'm trying to figure out what's different between loading the full session (which makes all of the same event calls and then all of the same buffer calls) and having the events coming in more slowly.
> 
> It may be that we need to batch up changes but I'd really like to know which limit we're hitting to make sure we fix it.

sudo-tee:
> I spent more time looking at this and I think it's related to treesitter. When able, can you try just commenting out the treesitter register line:
> 
> https://github.com/sudo-tee/opencode.nvim/blob/e93708de4e1e25fa326193a940cb1f0a45dcd257/lua/opencode/ui/ui.lua#L86

Disabling tressitter makes the buffer fly... like in the Replay mode without the full config. 

> If that works, can you comment that line back in but try turning off treesitter at the start of `renderer.on_part_updated` via `vim.treesitter.stop(state.windows.output_buf)` and then calling `vim.treesitter.start(state.windows.output_buf)` in trigger_on_data_rendered? That will debounce the async treesitter processing. The experience isn't great because the colors disappear but I just want to see if that actually makes the perf issue go away.

Disabling it on part_render, makes the same effect, the CPU never spike over 10%.

> I kind of wonder if the faster we make the event processing, the worse it is because that just queues up more work for treesitter and eventually we hit some limit. That would also make sense why doing `ReplayFullSession` or `ReplayAll 0` doesn't exhibit the problem because we do all of our work in a tight loop (without any vim.schedules/defer_fn) so the main loop doesn't have a chance to get treesitter looking at each of our tiny changes.

This was my intuition all along also.  There is to many changes at the same time so probably treesitter is scheduling some work and starts it... then another changes happens and it starts all over again. With the full session there is no time period available for treesitter to schedule some work.


> Assuming that's it, I don't think it's a good experience to turn off treesitter while we're waiting for more updates. And since I don't know of another way to control it, I think we'll have to do some kind of throttling. I have a first pass at the code but I'm still figuring out a good way to test it. Should have something tomorrow.

The experience of disabling it on each part render, is not good I agree, but it still better then having the ui locking up. (Maybe we can offer a config for people, if they have issues).

I played also a with throttling yestarday, in my tests it was really helping, but it caused way to many ui issues with text written at the wrong place or with the wrong extmarks, so I scraped the idea, because I didn't want to rewrite the whole rendered.

But this led me to an idea for a future refactoring.

Having a `Line` object with it's extmark + actions instead of having to keep a seperate lists in sync. This way it would be much easier to batch updates. But this is for the future.

Thanks again for your effort.

cameronr:
Great, glad we can confirm it's treesitter.

I think I have a clean way to throttle/batch (and even collapse overlapping) events at the EventManager level so I'm going to give that a shot.

sudo-tee:
You are right, this is probably the safest way to do it at the event manager level.

This way collapsing would be easier.

cameronr:
@sudo-tee Ok, just committed my throttling fix (it doesn't include any collapsing at the moment but it's possible it's not even necessary with the batching... but it'll be straightforward to add if we want it).

The basic idea is that we collect events in `ThrottlingEmitter` and every `drain_interval_ms`, it drains it's queue and hands a list of events back to `EventManager`, which means now we can process a batch of events all at once.

That is similar to how `ReplayAll 0` or `ReplayFullSession` work. Because we're not doing any vim.schedule in the processing, nvim/treesitter doesn't have a chance to get over-triggered.

As an additional benefit, I added a custom event to indicate the end of draining which we can use to auto scroll so I was able to clean up the scrolling logic a bit.

I also started using `EventManager` in the replay tool (instead of doing the dispatch manually) which is both better for testing (and needed now because I needed to test `ThrottlingEmitter`) and some nice code sharing.

Let me know how it works for you.

cameronr:
I added collapsing support. You can see it in action by running the replay tool with different values of `ReplayAll`.

`ReplayAll 0` will collapse as much as possible:

Screenshot 2025-10-25 at 12 52 52

A higher number, e.g. `ReplayAll 5`, will collapse significantly fewer since the events are streamed in and processed in batches:

Screenshot 2025-10-25 at 12 53 26

And a high enough number ( > than `event_throttle_ms`, default 40ms) will dispatch individual events

sudo-tee:
I just did some testing and I can say that it's now reeeally fast :)  

This is amazing.

I think with this I can safely merge. 

Let me know if you still need to look at things before.

I will do some last testing either tonight or tomorrow.

Amazing work

Thanks a lot for this.

cameronr:
I have a little bit of debug logging to remove. After that, we should just give a few days of bake time to make sure I didn't break anything / no other issues pop up. Will be great to get it out!

I'm soooo glad we had the unit tests and replay system. It wouldn't have been possible to improve things so quickly without them!

sudo-tee:
On a different note.

I am thinking of changing the plugin name and maybe soon having release numbers. GitHub supports renaming project so it should be fine

There is already another plugin named opencode.nvim.

I have a couple of names and wanted your input.

- openkode.nvim
- nvim-opencode
- neocoder.nvim
- opencoder.nvim
- opencode-ui.nvim
- codepal.nvim
- libre-code.nvim

I'm open for ideas ⁉️

cameronr:
Oh funny, I didn't even know about the other one. Native neovim UI and keymaps/navigation is such an improvement over a terminal integration, at least in my opinion... but I'm biased :)

On names, here are some that stand out to me:
- opencode.nvim (leave it as is and let the community sort it out)
- opencode-native.nvim (highlight the native aspect of it)
- neopencode.nvim (it's a little cumbersome but i think maintaining the connection to opencode is important)
- neo-opencode (another take on the idea above).

I do think keeping a direct connection to opencode is important from a "branding" perspective.

cameronr:
@sudo-tee It's purely cosmetic, but what do you think about switching the default thinking animation to:

Kapture 2025-10-26 at 08 23 53

The one above is smoother/less jumpy/distracting than the current default. If not, no worries.

sudo-tee:
> @sudo-tee It's purely cosmetic, but what do you think about switching the default thinking animation to:
> 
> Kapture 2025-10-26 at 08 23 53
> 
> The one above is smoother/less jumpy/distracting than the current default. If not, no worries.


Yes I'm ok with it. The current one can be distracting at time.

cameronr:
I've just been cleaning up some diagnostics. I did change the emmyrc to not include `$HOME/.local/share/nvim/lazy` as that can introduce some type conflicts if plugins define the same types (that was happening to me). I think it's probably the better choice but let me know if you think otherwise.

cameronr:
I did a bunch of testing and things look pretty good. I did find a few issues, summarized:

- clear all extmarks in output_window (needed because RenderMarkdown sometimes leaves extmarks behind) [6f424e0](https://github.com/sudo-tee/opencode.nvim/pull/62/commits/6f424e04f853fd28829c99f84fad2affa263258e)
- @ mentions were resetting the cursor position if you'd previously left the input window [4f1d172](https://github.com/sudo-tee/opencode.nvim/pull/62/commits/4f1d172429a66a34d910dc3b0a415934f0b90791)
- snacks file picker wasn't restoring the window, mode, cursor position [c1b9568](https://github.com/sudo-tee/opencode.nvim/pull/62/commits/c1b9568d187540730a3218e8c51fd13afe1e981f)
- report promise errors. this will surface problems like opencode config errors [e5cb4a1](https://github.com/sudo-tee/opencode.nvim/pull/62/commits/e5cb4a1b22f293fb7822b15866756babf82e686e)
- hopefully a permanent fix to spinner spinning at wrong times [097874c](https://github.com/sudo-tee/opencode.nvim/pull/62/commits/097874ca2be07588960e93bc28f60c25227921f2)
- more resilient aborting, including handling permissions [4b7187d](https://github.com/sudo-tee/opencode.nvim/pull/62/commits/4b7187dc2e25ea958a32e6c83ed9dc6f3277a2a3)

sudo-tee:
This is really nice. A bunch of little fixes. 

We are almost ready to merge.  

~Maybe one last thing.  I usually run with debug mode on. Can we have a setting for the events notifications under debug ?  So I can run without them ?~


After this I will probably merge as it looks stable to me also

EDIT:
My bad I was checking the wrong version

So if you feel it's safe to merge I will proceed.

sudo-tee:
@cameronr 

I just merged your amazing work.  🎉

Github didn't let me do a simple merge, only a squash would work. I did add you as a co-author because I din't want you to loose credit for all your work :).

Thank you again for your work  🔥

cameronr:
Oh great! I was just about to comment and say it looks good. Glad it's finally out!

PR #69, opened 2025-10-16
feat(context): context bar

sudo-tee:
This PR is adding a context bar to manage/view context being sent by the plugin


cameronr:
Ooh, I'm looking forward to this!

sudo-tee:
Yes I have some small issues after the rebase. But it's coming together.

sudo-tee:
@cameronr 

Whenever you have time you can have a look ?

You can review the code if you like and also tell me if the feature works for you.

I think it's going to be a nice feature but I might be biased :)

cameronr:
Yes, I'm very excited for this feature! will test and review today.

cameronr:
I haven't looked at the code yet but, for this kind of feature, I like to start with a review of the design / functionality

### Design:
- Overall, it looks nice :)
- The # popup menu is great
- Given how tight the input_window area is on space, I think it's worth asking two questions:
  1. Is the gear icon necessary?
  2. The extra line between the context bar and the prompt looks nice but it might be sacrificing too much vertical space. If we right align the content, maybe we could get away without needing that extra line? Here's what that would look like (at least with my config):

Screenshot 2025-10-30 at 13 47 42

### Functionality:

- Pressing #, highlighting an item, and then canceling selection / closing the blink autocomplete window, leaves the cursor on the line above where prompt input starts:

Kapture 2025-10-30 at 13 22 07

- The lines seem to behave strangely when pressing enter. I think it's related to that extra padding line? Most noticeable with `set nu` in the input window. 

Kapture 2025-10-30 at 13 51 09

It seems like that extra padding line is creating some interaction issues so I think the big question is if it's really necessary?

sudo-tee:
Thanks for the great feedback.

You are right about the padding. I added it for looks because it looked weird with the winbar ( mostly my frontend dev  background though it looked better 😅).  I like the suggestion to align it to the right. I will see what I can do about it.

I also saw some quirky behaviour with the padded line.  I will remove it and see if it feels better.

Don't hesitate if you have other ideas to improve it.

Thanks for your review.

lua/opencode/ui/completion/engines/blink_cmp.lua
```diff
@@ -55,12 +54,20 @@ function Source:get_completions(ctx, callback)
     for i, item in ipairs(source_items) do
       table.insert(items, {
         label = item.label,
-        kind = item.kind == 'file' and 17 or item.kind == 'folder' and 19 or 1, -- 17: File, 19: Folder, 1: Text
+        kind = item.kind,
+        kind_icon = item.kind_icon,
```
cameronr, line 58: If we set kind_hl, we can differentiate enabled vs not:

Screenshot 2025-10-30 at 16 36 25

For that, I just added this:

```lua
    kind_hl = not icon and available and 'WarningMsg' or nil,
```

to:
https://github.com/sudo-tee/opencode.nvim/blob/ff0ba7f5250cd37e1d8a3d2c8698d247f2de7ba0/lua/opencode/ui/completion/context.lua#L24-L28

And then just `kind_hl = item.kind_hl` in code this comment is linked to.

lua/opencode/ui/completion/engines/nvim_cmp.lua
```diff
@@ -50,18 +48,25 @@ function M.setup(completion_sources)
...
       local source_items = completion_source.complete(context)
       for j, item in ipairs(source_items) do
         table.insert(items, {
           label = item.label,
-          kind = item.kind == 'file' and cmp.lsp.CompletionItemKind.File
-            or item.kind == 'folder' and cmp.lsp.CompletionItemKind.Folder
-            or cmp.lsp.CompletionItemKind.Text,
+          kind = 1,
+          cmp = {
+            kind_text = item.kind_icon,
```
cameronr, line 58: Similar to blink, for cmp we can add this after line 58:

```lua
            kind_hl_group = item.kind_hl,
```

Image

lua/opencode/ui/completion/context.lua
```diff
@@ -0,0 +1,300 @@
...
+  local text_start = col - #completion_text
+
+  if text_start >= 0 and line:sub(text_start + 1, col) == completion_text then
+    line = line:sub(1, text_start) .. line:sub(col + 1)
+    input_win.set_current_line(line)
+    vim.api.nvim_win_set_cursor(0, { vim.api.nvim_win_get_cursor(0)[1], text_start })
+  elseif col > 0 and line:sub(col, col) == key then
+    line = line:sub(1, col - 1) .. line:sub(col + 1)
+    input_win.set_current_line(line)
+  end
```
cameronr, line 223: emmylua_ls suggests a nil check on col (and maybe this is a slightly cleaner way to get col):

```lua
  local col = vim.api.nvim_win_get_cursor(0)[2]

  if not col then
    return
  end
```

lua/opencode/ui/input_window.lua
```diff
@@ -191,6 +198,37 @@ function M.set_content(text, windows)
...
+function M.remove_mention(mention_name, windows)
+  windows = windows or state.windows
+  if not M.mounted(windows) then
+    return
+  end
+
+  local lines = vim.api.nvim_buf_get_lines(windows.input_buf, 0, -1, false)
+  for i, line in ipairs(lines) do
+    local mention_key = config.get_key_for_function('input_window', 'mention')
+    local updated_line = line:gsub(mention_key .. mention_name, '')
```
cameronr: Need to use vim.pesc here in case `mention_name` contains `-` (or other special chars)

```lua
    local pattern = vim.pesc(mention_key .. mention_name)
    local updated_line = line:gsub(pattern, '')
```

lua/opencode/config.lua
```diff
@@ -163,6 +165,9 @@ M.defaults = {
     selection = {
       enabled = true,
     },
+    agents = {
+      enabled = true,
+    },
```
cameronr, line 170: emmylua_ls complains that `subagents` field is missing (it was added to types later). Should be added here or marked as optional

lua/opencode/context.lua
```diff
@@ -1,53 +1,80 @@
 -- Gathers editor context
 
 local util = require('opencode.util')
-local config = require('opencode.config')
+local config = require('opencode.config') --[[@as OpencodeConfig]]
```
cameronr: Rather than setting the type here, maybe we should change the bottom of config.lua to be:

```lua
return setmetatable(M, {
  __index = function(_, key)
    return M.values[key]
  end,
  __newindex = function(_, key, value)
    M.values[key] = value
  end,
  __tostring = function(_)
    return vim.inspect(M.values)
  end,
}) --[[@as OpencodeConfig]]
```

lua/opencode/context.lua
```diff
@@ -292,9 +347,16 @@ local function format_diagnostics_part(diagnostics)
...
+  local buf = (M.get_current_buf() or 0) --[[@as integer]]
+  local lang = util.get_markdown_filetype(vim.api.nvim_buf_get_name(buf)) or ''
   return {
     type = 'text',
-    text = vim.json.encode({ context_type = 'cursor-data', line = cursor_data.line, column = cursor_data.column }),
+    text = vim.json.encode({
+      context_type = 'cursor-data',
+      line = cursor_data.line,
+      column = cursor_data.column,
+      line_content = string.format('`````%s\n%s\n`````', lang, cursor_data.line_content),
```
cameronr, line 362: yay, long codefences! :)

lua/opencode/context.lua
```diff
@@ -425,12 +485,39 @@ function M.extract_legacy_tag(tag, text)
...
+  vim.api.nvim_create_autocmd('DiagnosticChanged', {
+    pattern = '*',
+    callback = function(args)
+      local buf = args.buf
+      local curr_buf = state.current_code_buf or vim.api.nvim_get_current_buf()
+      if buf == curr_buf and util.is_buf_a_file(buf) and M.is_context_enabled('diagnostics') then
+        M.load()
+      end
+    end,
+  })
```
cameronr, line 527: could consider not setting up these autocmds if context is disabled?

lua/opencode/reactive_state.lua
```diff
@@ -0,0 +1,140 @@
+local M = {}
```
cameronr: I didn't see this file referenced anywhere?

lua/opencode/state.lua
```diff
@@ -39,6 +42,8 @@ local config = require('opencode.config')
 ---@field append fun( key:string, value:any)
 ---@field remove fun( key:string, idx:number)
 ---@field subscribe fun( key:string|nil, cb:fun(key:string, new_val:any, old_val:any))
+---@field opencode_server_job OpencodeServer
```
cameronr: This looks unreferenced?

lua/opencode/state.lua
```diff
@@ -52,11 +57,14 @@ local _state = {
...
   last_output_window_position = nil,
   last_code_win_before_opencode = nil,
+  current_code_buf = nil,
   display_route = nil,
   current_mode = config.default_mode,
   last_output = 0,
   -- context
   last_sent_context = nil,
+  current_context_config = {},
+  context_updated_at = nil,
```
cameronr, line 66: I only see `context_updated_at` written to. Is it actually used/needed?

lua/opencode/types.lua
```diff
@@ -53,7 +53,8 @@
 ---@field share? SessionShareInfo
 
 ---@class OpencodeKeymapEntry
----@field [1] string # Function name
+---@field [1] string # Function nam
```
cameronr: Accidental deletion of `e`?

lua/opencode/ui/completion/engines/blink_cmp.lua
sudo-tee: Oh I like that.  thanks

lua/opencode/config.lua
sudo-tee: It should just be agents, it was an error on my part

lua/opencode/context.lua
sudo-tee: I see your point, but if you toggle it later with something like 

Opencode run context=enabled=true  hello there 

you might end up with some stale context.

lua/opencode/reactive_state.lua
sudo-tee: Good catch. It was a failed attempt at having a reusable pattern for reactive state.  As the current one is not working for nested objects. I forgot to remove the file

lua/opencode/state.lua
sudo-tee: The context_bar is listening to it
image

This related to the reactive_state file.  I wanted to wrap the context to be able to subscribe to individual fields, but it ended up more complex than just having a timestamp. It's a bit of an anti-pattern but LuaJit does not support `ipairs`  in memtables so we cannot loop over indexed arrays when wrapped in memtables. 

Maybe I will find a better pattern later, but this one works for now and is simple enough

lua/opencode/types.lua
sudo-tee: Good catch it was indeed a mistake

sudo-tee:
@cameronr  

I did align the bar to the right and removed padding, It was a really great suggestion. It seems to work really well. I didn't encounter cursor jumps.

Let me know what you think.

There is still a couple of stuff I want to add like un the user message show that diagnostics have been shared like with section and now cursor data, a part from that I'm almost done

cameronr:
I took a quick look and it looks great!

My only other thought was if we should move the guard icon to be left aligned in the context bar? I think that would make more sense and probably look good?

cameronr:
Actually, now that I think about it, I wonder if the agent mode should be there too? It's not as clear (input guard is clearly input related) where as agent mode is more output related. That said, making sure the mode is set correctly before submitting is important and reducing the visual distance to verify that could be an improvement. Maybe it's worth at least looking at?

sudo-tee:
Those are great suggestions.

Current Model is also something related to input, but this could make the bar a little to crowded

I will play around with that idea.

sudo-tee:
I moved the prompt guard into the context-bar and you are right it make more sense there

image

sudo-tee:
Not convinced about the mode though. It looks a bit odd to me
image

Maybe it could be a future task to find a better place for the `model` and `mode`

cameronr:
Yeah, mode may be a stretch. My last idea is trying mode on the left rather than the right. My thinking is that from left to right it's conceptually:

- can i submit? (prompt guard, and, most of the time, it's not there)
- what mode am i submitting in (build/plan)
- what context am i including?

but it's possible that mode on the left just won't look good :)

if it doesn't look good, we can leave it up in the output_window winbar for now.

sudo-tee:
image

It could work..  but I think it's to punchy and to close to the content.

I will leave it at the top, until I find a better way to display it.  Maybe the status bar which is almost empty might be a good place in the future

PR #78, opened 2025-10-24
feat: add prompt_guard callback mechanism

guillaumeboehm:
This PR adds a plugin option to run a callback before sending promts to the LLM or before opening the opencode window.

In my use case I want to be able to exclude some directories of my computer with sensitive information in order to not mistakenly send them to the LLM.
I've mostly vibe-coded this and then went back on the code to check what it did, I'm not super well versed in lua so I might have missed things.

Implementation details:

- Add prompt_guard configuration option (function that returns boolean)
- Check guard before sending prompts (ERROR notification if denied)
- Check guard before opening buffer first time (WARN notification if denied)
- Add util.check_prompt_allowed() helper functions
- Guard has no parameters, users can access vim state directly
- Proper error handling for guard callback failures

### DIFF
--- a/lua/opencode/ui/output_window.lua
+++ b/lua/opencode/ui/output_window.lua
@@ -1,7 +1,5 @@
 local state = require('opencode.state')
 local config = require('opencode.config')
-local Timer = require('opencode.ui.timer')
-local topbar = require('opencode.ui.topbar')
 
 local M = {}
 M.namespace = vim.api.nvim_create_namespace('opencode_output')
@@ -157,36 +155,6 @@ function M.setup_keymaps(windows)
 end
 
 function M.setup_autocmds(windows, group)
-  local guard_timer = nil
-
-  local function start_guard_timer()
-    if guard_timer then
-      return -- Timer already running
-    end
-    guard_timer = Timer.new({
-      interval = 1000, -- 1 second
-      repeat_timer = true,
-      on_tick = function()
-        if state.windows and vim.api.nvim_win_is_valid(state.windows.output_win) then
-          topbar.render()
-          return true
-        end
-        return false
-      end,
-    })
-    guard_timer:start_and_tick()
-  end
-
-  local function stop_guard_timer()
-    if guard_timer then
-      guard_timer:stop()
-      guard_timer = nil
-    end
-  end
-
-  -- Start the timer when the window is created (visible) and tick immediately
-  start_guard_timer()
-
   vim.api.nvim_create_autocmd('WinEnter', {
     group = group,
     buffer = windows.output_buf,
@@ -204,15 +172,6 @@ function M.setup_autocmds(windows, group)
       vim.cmd('stopinsert')
       state.last_focused_opencode_window = 'output'
       require('opencode.ui.input_window').refresh_placeholder(state.windows)
-    end,
-  })
-
-  -- Stop the timer when the window is closed
-  vim.api.nvim_create_autocmd('WinClosed', {
-    group = group,
-    pattern = tostring(windows.output_win),
-    callback = function()
-      stop_guard_timer()
     end,
   })
 
--- a/lua/opencode/ui/prompt_guard_indicator.lua
+++ b/lua/opencode/ui/prompt_guard_indicator.lua
@@ -1,6 +1,5 @@
 local M = {}
 
-local state = require('opencode.state')
 local config = require('opencode.config')
 local util = require('opencode.util')
 local icons = require('opencode.ui.icons')
--- a/lua/opencode/ui/topbar.lua
+++ b/lua/opencode/ui/topbar.lua
@@ -137,6 +137,7 @@ function M.setup()
   state.subscribe('current_mode', on_change)
   state.subscribe('current_model', on_change)
   state.subscribe('active_session', on_change)
+  state.subscribe('is_opencode_focused', on_change)
   M.render()
 end
 
@@ -144,5 +145,6 @@ function M.close()
   state.unsubscribe('current_mode', on_change)
   state.unsubscribe('current_model', on_change)
   state.unsubscribe('active_session', on_change)
+  state.unsubscribe('is_opencode_focused', on_change)
 end
 return M


sudo-tee:
Thanks for the PR.

I like the idea, it gives a lot of flexibility.


I will have a look at the PR and come back to you

guillaumeboehm:
Sounds good

lua/opencode/core.lua
```diff
@@ -51,6 +51,14 @@ function M.open(opts)
   local are_windows_closed = state.windows == nil
 
   if are_windows_closed then
+    -- Check if opening buffer is allowed
+    local allowed, err_msg = util.check_prompt_allowed(config.prompt_guard)
```
sudo-tee: Not sure I get why it's needed here ?

It would prevent from opening the plugin.

README.md
```diff
@@ -533,6 +534,31 @@ The plugin defines several highlight groups that can be customized to match your
...
+    -- Return true to allow, false to deny
+    return true
+  end,
+})
+```
+
+### Behavior
+
+- **Before sending prompts**: The guard is checked before any prompt is sent to the AI. If denied, an ERROR notification is shown and the prompt is not sent.
+- **Before opening UI**: The guard is checked when opening the Opencode buffer for the first time. If denied, a WARN notification is shown and the UI is not opened.
```
sudo-tee, line 562: What is the usecase for preventing the ui to open ?

lua/opencode/util.lua
```diff
@@ -347,4 +347,30 @@ function M.parse_dot_args(args_str)
   return result
 end
 
+--- Check if prompt is allowed via guard callback
+--- @param guard_callback? function
+--- @return boolean allowed
+--- @return string|nil error_message
+function M.check_prompt_allowed(guard_callback)
```
sudo-tee: I would also allow to modify the text of the prompt ?

Even the context object where you can remove files or anything.

It would be more flexible. If the function see a file in the context it should not sent it could call context.remove_file (the function is not there yet but would be trivial to add).

What do you think ?

README.md
guillaumeboehm: I felt like letting the user go all the way until the prompt, the file selection and everything is all done only to then show an error when sending would be quite frustrating ^^' Maybe the window could open normally with a warning notification stating that prompt guard is enabled ?

lua/opencode/util.lua
guillaumeboehm: Being able to manipulate everything before prompting is ok I think yes, but automatic removal of the files feels a little useless to me... Because the AI can still ask to read files outside of the files selected by this plugin I'm not sure it's very useful, this would be more of the role of some mcp or tool restricting filesystem access no? Although I think that giving the list of selected files as a callback parameter could be interesting I suppose.

In my usecase I just look at the cwd to know if we're authorized to make AI calls or not.

README.md
sudo-tee: This would be a better option than to not open the plugin.

guillaumeboehm:
I'm trying some stuff with the mentioned files, so I'm marking this as draft for now.

guillaumeboehm:
Alright, I think the PR is ready.
I've changed the error on open with a warning and added a little indicator on the left of the topbar shown when the guard is going to deny prompts.
I've also added the mentioned files list to the prompt_guard callback so the user can make a decision based on that, I didn't add any smart logic, as I felt it would be a bit half effective as I explained earlier.

lua/opencode/ui/output_window.lua
```diff
@@ -155,6 +157,36 @@ function M.setup_keymaps(windows)
 end
 
 function M.setup_autocmds(windows, group)
+  local guard_timer = nil
+
+  local function start_guard_timer()
+    if guard_timer then
+      return -- Timer already running
+    end
+    guard_timer = Timer.new({
```
sudo-tee: Is the poupose of the timer only to check the gard status ?

Whan can then guard change ? When files are added ?

If it's the case then having an event being sent whan the context files changes, would be better.

Then the topbar could listen to that event ?

```lua

-- In context.lua add files
state.event_manager:emit('custom.context.file.added', {file = file})


--- in topbar
state.event_manager:subscribe('custom.context.file.added', M.render)
```

Having a timer check a user function every second could lead to bad performance if the function do blocking things.

The worst case would be an outdated topbar. but since the guard is checked again before sending the message it would not be that bad ?

What do you think ?

lua/opencode/ui/prompt_guard_indicator.lua
```diff
@@ -0,0 +1,37 @@
...
+---Get formatted indicator string with highlight (empty if allowed)
+---@return string formatted_indicator
+function M.get_formatted()
+  if not M.is_denied() then
+    -- Prompts are allowed - don't show anything
+    return ''
+  end
+
+  -- Prompts will be denied - show red indicator
+  local icon = icons.get('status_off')
```
sudo-tee: I would put a shield icon.. 

status_off is confusing  since it's a switch

I would prefer something like this 
Image

You search for shield here.
https://www.nerdfonts.com/cheat-sheet

lua/opencode/ui/topbar.lua
```diff
@@ -40,17 +42,45 @@ local function get_mode_highlight()
...
+local function create_winbar_text(description, model_info, mode_info, show_guard_indicator, win_width)
+  -- Calculate how many visible characters we have
+  -- Format: " [GUARD] description padding model_info MODE "
+  -- Where [GUARD] is optional (1 char + 1 space = 2 visible chars)
+  
+  local guard_prefix = ''
+  local guard_visible_width = 0
+  
+  if show_guard_indicator then
+    local guard_icon = icons.get('status_off')
```
sudo-tee: Why not use the `get_formatted` function from the prompt_guard file ?

sudo-tee (changes requested):
I like this better, and the icon in the topbar is a nice touch.

However I would avoid having a timer running. A simple event subscription would be cleaner.

lua/opencode/ui/prompt_guard_indicator.lua
guillaumeboehm: Yeps looks good like this you're right

lua/opencode/ui/topbar.lua
guillaumeboehm: Erf you're right, AI goodness that I missed.

lua/opencode/ui/output_window.lua
guillaumeboehm: Hum... Yeah I see the issues it could cause. The idea was that whatever is used inside the guard callback could change outside of the control of this plugin, on some event, and I felt like keeping the state up to date was important.
But I guess maybe it's ok if it stays outdated for a bit?
Maybe an event when focusing the input window? Although we could end up with the same issue of some lag when focusing the window... I don't know if updating the status when trying to send a message would be enough... What do you think?

sudo-tee: I think we should start small and simple.  

You can subscribe to state.is_opencode_focused as well



If it becomes an issue that the toolbar is not in sync we can always address it later.

guillaumeboehm:
Ok, I think I understood what you meant with the is_focused subscription. I think it's a pretty good compromise actually.

sudo-tee:
Thanks for your effort on the PR, I just merged it

PR #80, opened 2025-10-27
Respect user set default_mode and improve agent management 

guillaumeboehm:
This PR aims to improve some of the agent switching and default behaviors.

I encountered an issue because my default agents are disabled in my opencode config, and the plugin always added them to the available agents and ignored the default_mode set by the user.

Here's an overview of what this PR changes:

- The default_mode option was not being respected when setting the
  state current_mode. The current_mode is now set when first opening the
  opencode window.
- The mode switching is now wrapped in a "switch_to_mode" function to
  ensure the mode exists in the available agents for more control.
- The build and plan agents were added to the available agents even if
  explicitly disabled in the opencode config -> leading to a hang when
  sending messages to opencode. They are now prepended if they are not
  disabled.
- The agent list was always randomly returned, it is now sorted before
  using it.

### DIFF
--- a/tests/unit/config_file_spec.lua
+++ b/tests/unit/config_file_spec.lua
@@ -64,6 +64,21 @@ describe('config_file.setup', function()
     assert.True(vim.tbl_contains(agents, 'plan'))
   end)
 
+  it('get_opencode_agents respects disabled defaults', function()
+    state.api_client = {
+      get_config = function()
+        return Promise.new():resolve({ agent = { ['custom'] = { mode = 'primary' }, ['build'] = { disable = true }, ['plan'] = { disable = false } } })
+      end,
+      get_current_project = function()
+        return Promise.new():resolve({ id = 'p1' })
+      end,
+    }
+    local agents = config_file.get_opencode_agents()
+    assert.True(vim.tbl_contains(agents, 'custom'))
+    assert.False(vim.tbl_contains(agents, 'build'))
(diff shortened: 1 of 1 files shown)


guillaumeboehm:
Ah looks like the tests are broken... I obviously ignored them as hard as I could 🤣 I'll figure out how to run them and check what's wrong.

sudo-tee:
Take your time. Otherwise it looks like a good feature

guillaumeboehm:
> Take your time. Otherwise it looks like a good feature

Yeah I figured out what's the issue but I'm struggling to fix it xD When the renderer test is being run, it's using the user opencode config instead of a mock config, and the expected 'BUILD' agent in the topbar is not here in my case.
So I'm trying to make the renderer tests use a mock config, but it's not going smooth xD

sudo-tee:
@guillaumeboehm 

I've had a look at it and I think it's just that the mode is never set for the renderer_spec as it never calls api.open that calls your ensure_mode. Where before the mode was set directly in the state.

I pushed a fix for the tests to your branch.

I also removed a render_output in the ensure_mode, as it's unecessary since it's done in the open method.

Can you just do a last test to ensure everthing is fine, and after that I can merge this

guillaumeboehm:
Thanks! It works great now I think. I also added a little test in the config_file tests to test the disabled agents case.

PR #83, opened 2025-10-28
fix(renderer): occasional parts rendering, markdown types

cameronr:
This started out as a smaller bugfix but along the way I noticed that sometimes parts were not rendering, so that's the more important issue, from: [4317d0b](https://github.com/sudo-tee/opencode.nvim/pull/83/commits/4317d0bc61c1b47ac9ac54419adddbba4752e1de):

This was a subtle one and it only shows up when we're not collapsing
events so it wasn't caught by the unit tests. I've since modified the
unit tests to also run the replays without collapsing to catch bugs like
this in the future.

If the first version of a part didn't result in something being
rendered, future updates to that part would also not render anything.
This was particularly likely to happen with tools as the first version
of the part is often `step-start`. Because we didn't think this was a
new part, we called `_replace_part_in_buffer` but that exits quickly
because the part hadn't been rendered anywhere.

While we could try to call `_insert_part_to_buffer` in
`_replace_part_in_buffer` if the part hadn't been rendered, that would
miss some subtle cases around error handling so the correct fix is
consider the part new in `on_part_updated` if we haven't rendered it
before.

Now, onto the issue that started this PR:

I noticed that embedded codefences were screwing up our rendering. At first, I thought we needed to escape/replace them but then I realized we could use a longer codefence for our outer wrapper. I also realized that we weren't setting the markdown types correctly (markdown should be `markdown` not `md`). Setting the codefence to `markdown` enables render-markdown to render embedded codefences correctly.

Before:

Screenshot 2025-10-28 at 16 16 24

After (the codefence is normally hidden, I just have the cursor there to show the longer codefence and correct type)

Screenshot 2025-10-28 at 21 10 55

### DIFF
--- a/lua/opencode/util.lua
+++ b/lua/opencode/util.lua
@@ -385,6 +385,32 @@ function M.get_markdown_filetype(filename)
   end
 
   return vim.fn.fnamemodify(filename, ':e')
+--- Check if prompt is allowed via guard callback
+--- @param guard_callback? function
+--- @param mentioned_files? string[] List of mentioned files in the context
+--- @return boolean allowed
+--- @return string|nil error_message
+function M.check_prompt_allowed(guard_callback, mentioned_files)
+  if not guard_callback then
+    return true, nil -- No guard = always allowed
+  end
+
+  if not type(guard_callback) == 'function' then
+    return false, 'prompt_guard must be a function'
+  end
+
+  mentioned_files = mentioned_files or {}
+  local success, result = pcall(guard_callback, mentioned_files)
+
+  if not success then
+    return false, 'prompt_guard error: ' .. tostring(result)
+  end
+
+  if type(result) ~= 'boolean' then
+    return false, 'prompt_guard must return a boolean'
+  end
+
+  return result, nil
 end
 
 return M
--- a/README.md
+++ b/README.md
@@ -240,6 +240,7 @@ require('opencode').setup({
   debug = {
     enabled = false, -- Enable debug messages in the output window
   },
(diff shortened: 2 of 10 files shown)


cameronr:
@sudo-tee This one is probably a little urgent so would love a review when you have time.

lua/opencode/util.lua
```diff
@@ -143,9 +143,10 @@ end
 ---Strip ANSI escape sequences from all lines
 ---@param lines table
 ---@return table stripped_lines
-function M.strip_ansi_lines(lines)
+function M.sanitize_lines(lines)
   local stripped_lines = {}
   for _, line in pairs(lines) do
+    -- vim.notify(line)
```
sudo-tee: Rogue comment left out :)

sudo-tee (approved):
This is a great fix.

Thanks you for this

There is a small conflict with main, once it's resolved I will merge

sudo-tee:
@cameronr 

I'm sorry I think I messed up the conflict resolution with the github UI.

cameronr:
no worries, i can take a look in a few hours. fwiw, the GH UI merges have let me down before so I've stopped doing merges there.

also, if you'd rather i can just rebase my changes on main and resubmit the pr?

sudo-tee:
> no worries, i can take a look in a few hours. fwiw, the GH UI merges have let me down before so I've stopped doing merges there.
> 
> also, if you'd rather i can just rebase my changes on main and resubmit the pr?

Yes that would probably be easier to just rebase and drop the merge commit I did

PR #84, opened 2025-10-29
refactor(cmds): move all cmds under Opencode

### LINKED ISSUE #82: Feature: All cmds should be under a single Opencode cmd
As discussed in #43, all `:` cmds should be under a single `Opencode` cmd to keep the cmd namespace clean.

Here's my proposal for the commands:

**Structure:**
```vim
:Opencode [subcommand] [args]
```

**Proposed Subcommands:**

```vim
:Opencode " Toggle (default action)
:Opencode open [input|output]
:Opencode close
:Opencode cancel             " Renamed from stop, cancels current request, if there is one
:Opencode toggle             " Toggle open/close
:Opencode toggle_focus        " Toggle focus between opencode and code window
:Opencode toggle_pane         " Toggle between input/output panes
:Opencode swap

" Session Management
:Opencode session new [title]
:Opencode session select
:Opencode session child
:Opencode session compact
:Opencode session share
:Opencode session unshare
:Opencode session agents_init " Renamed from init, analyze app and create AGENTS.md

" History
:Opencode undo
:Opencode redo

" Diff/Review
:Opencode diff [open|next|prev|close]
:Opencode revert all [prompt|session|snapshot]
:Opencode revert this [prompt|session|snapshot]
:Opencode restore file <snapshot>
:Opencode restore all <snapshot>

" Agent/Mode
:Opencode agent [plan|build|select]
:Opencode models

" Utilities
:Opencode run <prompt>
:Opencode run_new <prompt>
:Opencode help
:Opencode mcp

" Permissions
:Opencode permission [accept|accept_all|deny]
```

Notes: use `snake_case` for multi-word sub-cmds and arguments.

## Open questions
- [x] Is backwards compatibility important? Yes, with a warning
- [ ] Are there any commands missing or any that we can remove?

cameronr:
Backwards compatibility maintained with config.legacy_commands option.
It's currently true but should be set to false at some point in the
future.

Fixes #82

### DIFF
--- a/lua/opencode/api.lua
+++ b/lua/opencode/api.lua
@@ -958,24 +958,24 @@ M.commands = {
   run = {
     desc = 'Run prompt in current session',
     fn = function(args)
-      local prompt = table.concat(args, ' ')
+      local opts, prompt = util.parse_run_args(args)
       if prompt == '' then
         vim.notify('Prompt required', vim.log.levels.ERROR)
         return
       end
-      M.run(prompt)
+      M.run(prompt, opts)
     end,
   },
 
   run_new = {
     desc = 'Run prompt in new session',
     fn = function(args)
-      local prompt = table.concat(args, ' ')
+      local opts, prompt = util.parse_run_args(args)
       if prompt == '' then
         vim.notify('Prompt required', vim.log.levels.ERROR)
         return
       end
-      M.run_new_session(prompt)
+      M.run_new_session(prompt, opts)
     end,
   },
 
--- a/lua/opencode/util.lua
+++ b/lua/opencode/util.lua
@@ -420,4 +420,38 @@ function M.strdisplaywidth(str)
   return vim.fn.strdisplaywidth(str)
 end
 
+--- Parse run command arguments with optional agent, model, and context prefixes.
+--- Returns opts table and remaining prompt string.
+--- Format: [agent=<name>] [model=<model>] [context=<key=value,...>] <prompt>
+--- @param args string[]
+--- @return table opts, string prompt
+function M.parse_run_args(args)
+  local opts = {}
+  local prompt_start_idx = 1
+
+  for i, token in ipairs(args) do
+    local agent = token:match('^agent=(.+)$')
+    local model = token:match('^model=(.+)$')
+    local context = token:match('^context=(.+)$')
+
+    if agent then
+      opts.agent = agent
+      prompt_start_idx = i + 1
+    elseif model then
+      opts.model = model
+      prompt_start_idx = i + 1
+    elseif context then
+      opts.context = M.parse_dot_args(context:gsub(',', ' '))
+      prompt_start_idx = i + 1
+    else
+      break
+    end
+  end
+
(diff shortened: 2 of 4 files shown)


sudo-tee:
Great work on this.

I found an issue with arguments, but to be honest they were made in a weird way

When I run `Opencode run_new hello context.current_file.enabled=false` it used to use parse the flags at the end

image

This is in the readme 

image


But now it's not working anymore they end up in the prompt.

image


They probably should start with `--` instead of relying on the first equal sign `something=this` but it would make backward compatibility harder.

cameronr:
Ooh, good catch. I'll take a look. Thanks!

cameronr:
I had a few mins to look at this. I agree; I really don't like plucking the extra arguments from the prompt. I propose we break backwards compatibility here (it probably isn't used very often anyway) and do the following:

#### Format
```
:OpencodeRun [agent=<name>] [model=<model>] [context=<overrides>] <prompt>
:OpencodeRunNewSession [agent=<name>] [model=<model>] [context=<overrides>] <prompt>
```

All three are optional, can appear in any order, but must come before the prompt.

#### Examples

**Single prefix:**
```vim
:OpencodeRun agent=plan what's the architecture here?
:OpencodeRunNewSession model=github-copilot/gpt-4.1 fix this bug
:OpencodeRun context=current_file.enabled=false analyze the project
```

**Multiple prefixes (any order):**
```vim
:OpencodeRun agent=plan model=github-copilot/claude-sonnet-4 review my code
:OpencodeRunNewSession model=anthropic/claude-3.5 agent=build context=selection.enabled=false refactor this
:OpencodeRun context=diagnostics.error=true,current_file.enabled=false agent=plan model=openai/gpt-4 what are the errors?
```

**No prefixes:**
```vim
:OpencodeRun Fix the bug in the current file
```

**Complex context with multiple overrides:**
```vim
:OpencodeRunNewSession agent=plan context=current_file.enabled=false,selection.enabled=true,diagnostics.warn=false help me understand this
```

#### Parsing Logic
1. Scan tokens from left to right
2. If token matches `agent=*`, extract agent name
3. If token matches `model=*`, extract model
4. If token matches `context=*`, pass the value to `parse_dot_args()`
5. Everything else (first non-matching token onward) becomes the prompt

#### Context value format
The `context=` value would be comma-delimited key=value pairs using dot notation, parsed by `parse_dot_args()`:
- `context=current_file.enabled=false`
- `context=current_file.enabled=false,diagnostics.info=true,selection.enabled=false`

How's that sound?

sudo-tee:
To.me it sounds pretty reasonable. 

I agree that by looking at usage on GitHub it does not seem to be a used feature at the moment.

So I would be ok with your solution.

cameronr:
Ok, I think this is all set. Whenever you have time, let me know if you see any other issues.

sudo-tee:
Looks good to me.

I think there is an issue where the context is not loaded properly whan using the context params, but it's not a bug in your MR.

This should be adressed in the context bar i'm doing.

Thanks a lot for your great PR

PR #86, opened 2025-10-30
fix(topbar): fix spacing

cameronr:
There was an off by one for the mode string.

Before:
Screenshot 2025-10-29 at 16 38 08
Screenshot 2025-10-29 at 16 38 14

After:
No guard, no truncation, plan
Screenshot 2025-10-29 at 17 55 55

No guard, truncation, plan
Screenshot 2025-10-29 at 17 56 02

No guard, truncation, build
Screenshot 2025-10-29 at 17 56 08

Guard, no truncation, plan
Screenshot 2025-10-29 at 17 55 22

Guard, truncation, plan
Screenshot 2025-10-29 at 17 55 33

Guard, truncation, build
Screenshot 2025-10-29 at 17 56 30

As a design choice, I extend the top bar all the way to the left and right (rather than leaving one empty space on both sides. Given how much we want to show in that area, I figured it was best to have as much room for guard, description, model, and mode

I also tried cleaning the code up to make it easier to think about and, hopefully, extend, if needed.

Not for this request, but I also wonder if the guard icon would make more sense to display in the input area? Maybe something for your context bar at some point.

### DIFF
--- a/lua/opencode/ui/topbar.lua
+++ b/lua/opencode/ui/topbar.lua
@@ -1,3 +1,5 @@
+local util = require('opencode.util')
+
 local M = {}
 
 local state = require('opencode.state')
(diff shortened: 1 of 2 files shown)


sudo-tee:
Nice catch.

I also start to think that the toolbar is getting a little bit out of control. 

I think that the model should be close to the input window in the future.  

The guard icon could also be moved.

PR #90, opened 2025-10-31
Add desc to default editor keymaps

mikaoelitiana:
This PR updates the default editor keymap to include a desc parameter for each key binding, providing a human-readable short action name. This improves discoverability and clarity for users configuring or inspecting keymaps.

This makes it look nicer in which-key 

image


cameronr:
@sudo-tee Looks good to me. Is it ok I approve smaller pull requests like this?

sudo-tee:
Yes sure . I'm ok with that

PR #92, opened 2025-11-01
fix(renderer): handle inserting out of order parts

cameronr:
To support sending multiple messages, the main thing that we needed to support was inserting parts not at the end. I didn't need to build a queue of outgoing messages because we can just send them to Opencode as we get them.

This PR also includes not using relative timestamps as they don't make sense when we're not re-rendering. The time should be formatted the way the Opencode TUI does it.

I've tested simultaneous messages a bit but it could definitely use another set of eyes.


sudo-tee:
Great. It was a lot easier than I anticipated.

Thanks for the PR

PR #96, opened 2025-11-04
Add current_model function to the API

guillaumeboehm:
I had a little need for something like this in my confs. And I felt like it could be a useful addition so here's a PR.

I'm still testing it a bit so I'll undraft it in a bit and after adding some unit tests.

### DIFF
--- a/tests/unit/api_spec.lua
+++ b/tests/unit/api_spec.lua
@@ -487,4 +487,35 @@ describe('opencode.api', function()
       config_file.get_user_commands = original_get_user_commands
     end)
   end)
+
+  describe('current_model', function()
+    it('returns the current model from state', function()
+      local original_model = state.current_model
+      state.current_model = 'testmodel'
+
+      local model = api.current_model()
+      assert.equal('testmodel', model)
+
+      state.current_model = original_model
+    end)
+
+    it('falls back to config file model when state.current_model is nil', function()
+      local original_model = state.current_model
+      state.current_model = nil
+
+      local config_file = require('opencode.config_file')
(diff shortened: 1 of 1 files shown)


cameronr:
@sudo-tee This specific change looks fine to me.

It does make me wonder, tho, if there's a better way to initialize `state.current_model`. Currently, it's kind of lazily set, usually via topbar:

https://github.com/sudo-tee/opencode.nvim/blob/c3ad99562f6bb8745a1c2304589a4d6c1dabadb1/lua/opencode/ui/topbar.lua#L12-L16

Also possible it gets set by renderer:
https://github.com/sudo-tee/opencode.nvim/blob/c3ad99562f6bb8745a1c2304589a4d6c1dabadb1/lua/opencode/ui/renderer.lua#L932-L935

Or it can be specifically selected:
https://github.com/sudo-tee/opencode.nvim/blob/c3ad99562f6bb8745a1c2304589a4d6c1dabadb1/lua/opencode/core.lua#L180-L189

Would it be better to explicitly initialize so it's always well defined? Or maybe have a way in `state` to specify a default initializer function on read when a value is nil?

sudo-tee:
You are right. I've been thinking about that lately since I got annoyed by the fact that after a  restart if you had a session with a specific model it defaults back to the default model instead of picking the one from the last message. 

The model should probably be set when the config is loaded if the config specifies one.  And should also be determined by the last message in an active session. But it's tricky because if a new session is created after, it should be reset to default IMHO.

A default initializer is a good idea, but only triggered once the config is loaded.

I hope it makes sense?

cameronr:
I'll take a stab at solving this.

PR #98, opened 2025-11-04
refactor: carefully set model from session

cameronr:
When loading a session, clear state.current_model and set it to the model used by the most recent message.

Topbar shouldn't set a model as a side effect, it should just display what's been selected.

Add `core.initialize_current_model()` which will initialize state.current_model to the default from the config_file if it hasn't been set

I think this achieves the behavior described in https://github.com/sudo-tee/opencode.nvim/pull/96#issuecomment-3487956371:

- If you load a session, it will set `state.current_model` to the model most recently used in that session
- If you create a new session, `state.current_model` will get re-initialized to the default
- Rerendering a session won't change `state.current_model`

### DIFF
--- a/lua/opencode/api.lua
+++ b/lua/opencode/api.lua
@@ -355,7 +355,7 @@ function M.initialize()
     vim.notify('Failed to create new session', vim.log.levels.ERROR)
     return
   end
-  if not state.current_model then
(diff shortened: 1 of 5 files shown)


sudo-tee:
Thanks for having a look at this.

I wanted to have a look pretty soon. You beat me to it 😁.

PR #99, opened 2025-11-04
fix(api): use core.initialize_current_model

cameronr:
Very minor, but makes sense to use the helper we just made. I forgot to mention it before #96 was merged.


sudo-tee:
I was too fast on merging the other PR.

My bad

cameronr:
I love how fast you and the project are moving :)

PR #100, opened 2025-11-05
fix(config_file): provider filtering 

cameronr:
Filter was broken so count percentages weren't showing in the footer

Also did some types/diagnostics cleanup.


sudo-tee:
Great cleanup, and good catch for the fix.

PR #101, opened 2025-11-05
feat(timeline-picker): implement basic timeline picker

sudo-tee:
This implement a simple timeline pcker similar to the one found in opencode tui
image

### DIFF
--- a/lua/opencode/ui/base_picker.lua
+++ b/lua/opencode/ui/base_picker.lua
@@ -360,7 +360,7 @@ end
 ---@param width? number Optional width override
 ---@return PickerItem
 function M.create_picker_item(text, time, debug_text, width)
-  local time_width = time and #util.format_time(0) + 1 or 0 -- longest time format by using 0
+  local time_width = time and #util.format_time(time) + 1 or 0
   local debug_width = config.debug.show_ids and debug_text and #debug_text + 1 or 0
   local item_width = width or vim.api.nvim_win_get_width(0)
   local text_width = item_width - (debug_width + time_width)


cameronr:
Oh nice! I'm happy to review when it's ready

sudo-tee:
@cameronr 

If you want to have a look, it's almost ready

cameronr:
Great, I'll take a look today

cameronr:
Had a few mins and tried to play around with it but ran into some early errors:

1. Getting an error when typing in the session picker (using snacks). I assume related to the picker_utils unification?
```
Unhandled async error:
...nvim/lazy/snacks.nvim/lua/snacks/picker/core/matcher.lua:524: attempt to index local 'str' (a nil value)
```

2. Error when I try to open the timeline:
```
E5108: Error executing lua: ...im-dev/opencode.nvim/lua/opencode/ui/timeline_picker.lua:4: module 'lua.opencode.config' not found:
```

Removing the "lua." does let me open the picker. I'll take a closer look but wanted to share those issues in case they were already on your radar.

cameronr:
Do you think `picker_util` can get the size of the picker window and right align the dates across all of the picker plugins?

Screenshot 2025-11-06 at 15 29 20
Screenshot 2025-11-06 at 15 30 14
Screenshot 2025-11-06 at 15 31 24

lua/opencode/ui/completion/commands.lua
```diff
@@ -76,7 +76,9 @@ local command_source = {
           vim.api.nvim_win_set_cursor(0, { 1, #item.insert_text + 1 })
           return
         end
-        item.data.fn()
+        vim.defer_fn(function()
+          item.data.fn()
+        end, 10) -- slight delay to allow completion menu to close, this prevent a weird bug with mini.pick where it displays an empty window with `BlinkDonotRepeatHack` text inserted
```
cameronr, line 81: I've seen things like that before. I'm curious if vim.schedule would've been enough or if it really requires deferring for a short time?

lua/opencode/ui/navigation.lua
```diff
@@ -20,6 +20,24 @@ local function is_message_header(details)
...
+
+  if not win or not buf then
+    return
+  end
+
+  local rendered_msg = require('opencode.ui.renderer').get_rendered_message(message_id)
+  if not rendered_msg or not rendered_msg.line_start then
+    return
+  end
+  local sep_offset = 2
```
cameronr: Maybe slightly better to count the number of lines in formatter.separator?

lua/opencode/core.lua
```diff
@@ -25,21 +25,26 @@ function M.select_session(parent_id)
...
+function M.switch_session(session_id)
+  local selected_session = session.get_by_id(session_id)
+  -- clear the model so it can be set by the session. If it doesn't get set
+  -- then core.get_model() will reset it to the default
+  state.current_model = nil
+  state.active_session = selected_session
+  if state.windows then
+    state.restore_points = {}
+    -- Don't need to update either renderer because they subscribe to
+    -- session changes
```
cameronr: Not important, but now we only have one renderer, could update the comment here :)

lua/opencode/ui/completion/commands.lua
sudo-tee: Vim schedule was not working unfortunately

sudo-tee:
> Had a few mins and tried to play around with it but ran into some early errors:
> 
> 1. Getting an error when typing in the session picker (using snacks). I assume related to the picker_utils unification?
> ```
> Unhandled async error:
> ...nvim/lazy/snacks.nvim/lua/snacks/picker/core/matcher.lua:524: attempt to index local 'str' (a nil value)
> ```
> 
> 2. Error when I try to open the timeline:
> ```
> E5108: Error executing lua: ...im-dev/opencode.nvim/lua/opencode/ui/timeline_picker.lua:4: module 'lua.opencode.config' not found:
> ```
> 
> Removing the "lua." does let me open the picker. I'll take a closer look but wanted to share those issues in case they were already on your radar.

I will soon have a refactored picker that both sessions and timeline can use. It will be much easier to create new ones

sudo-tee:
> It's coming along really nicely!
> 
> My only suggestion is that we should do some more work on how the pickers format their text:
> 
> 1. Dates should be right aligned consistently across the picker providers
> 2. I wonder if there's an opportunity for a more condensed date format, maybe it's "Oct 05 14:23" if it's this year, and "2024-10-05 14:23" if not?

Great idea. 

For the alignment. I need to find a way to get the picker width.

sudo-tee:
> It's coming along really nicely!
> 
> My only suggestion is that we should do some more work on how the pickers format their text:
> 
> 1. Dates should be right aligned consistently across the picker providers
> 2. I wonder if there's an opportunity for a more condensed date format, maybe it's "Oct 05 14:23" if it's this year, and "2024-10-05 14:23" if not?

@cameronr 

I did some work on having the dates align on all pickers.

Unfortunately some pickers construct their list before showing the window so getting the with was not possible.

I ended up fixing the with of the picker instead with a config, It works pretty well with some edges cases where telescope and fzf have somme padding in their default theme. I added a padding offset for theeses pickers.  It should be good enough.

sudo-tee:
Telescope
image

Snacks

image

FZF

image

mini.pick

image

cameronr:
With my font settings, the times/dates are completely hidden with snacks (but the others are fine):

Screenshot 2025-11-07 at 11 21 30

For snacks, the width is available in the format function, so instead wrapping `format_fn` and hardcoding `opts.width`, we could leave format function as is and then for pickers that have a width available, they can use that:

```lua
    format = function(item, snacks_picker)
      return opts.format_fn(item, snacks_picker.layout.width):to_formatted_text()
    end,
```

For other pickers, they could just pass `opts.width`

I made that change and here's what it looks like (other pickers are unchanged):

Screenshot 2025-11-07 at 11 37 11

I made two other changes:
- fixed snacks typing (it requires `item.text` to be set)
- used the actual time to set `time_width`. I know that means that the ... transition will happen differently for short times vs long times but I think that's actually more desirable (and it shouldn't be jarring because it'll be one time transition from short dates to long dates).

Hopefully those changes are ok!

cameronr:
Oh shoot, I just saw your update to set the snacks layout widths. I'll undo the the snacks width changes.

cameronr:
Ok, so now my changes are just fixing snacks matching and using the actual time for width calculation.

Everything else looks good. Really nicely done!

sudo-tee:
Thanks for your fixes. 

I will merge this pretty soon

PR #104, opened 2025-11-05
Improve scroll at bottom to support scrolling without focus

guillaumeboehm:
Took a shot at improving a bit the autoscroll logic, I'm using the mouse to scroll from time to time and the previous implementation was not supporting that.

I also added a config to always autoscroll, I felt like it could be useful. If you don't like it it's separated in its own commit.

### DIFF
--- a/lua/opencode/ui/output_window.lua
+++ b/lua/opencode/ui/output_window.lua
@@ -32,7 +32,6 @@ end
 ---@param win? integer Window ID, defaults to state.windows.output_win
 ---@return boolean true if at bottom, false otherwise
 function M.is_at_bottom(win)
-  -- If always_scroll_to_bottom is enabled, always return true
   if config.ui.output.always_scroll_to_bottom then
     return true
   end
@@ -40,7 +39,7 @@ function M.is_at_bottom(win)
   win = win or (state.windows and state.windows.output_win)
 
   if not win or not vim.api.nvim_win_is_valid(win) then
-    return true  -- Assume at bottom if window invalid
+    return true
   end
 
   if not state.windows or not state.windows.output_buf then
@@ -49,7 +48,7 @@ function M.is_at_bottom(win)
 
   local ok, line_count = pcall(vim.api.nvim_buf_line_count, state.windows.output_buf)
   if not ok or not line_count or line_count == 0 then
-    return true  -- Empty buffer, consider at bottom
+    return true
   end
 
   local botline = vim.fn.line('w$', win)
@@ -215,7 +214,6 @@ function M.setup_autocmds(windows, group)
     group = group,
     buffer = windows.output_buf,
     callback = function()
-      -- Update state to track if user is at bottom
       M.viewport_at_bottom = M.is_at_bottom(windows.output_win)
     end,
   })


cameronr:
Nice! I've also been a little annoyed when i've scrolled the view up (but not repositioned the cursor).

My only suggestion is that I don't think `output_window_at_bottom` needs to be in `state`. I think it would be fine to have something like `output_window.viewport_at_bottom` and renderer can just check that.

guillaumeboehm:
> Nice! I've also been a little annoyed when i've scrolled the view up (but not repositioned the cursor).
> 
> My only suggestion is that I don't think `output_window_at_bottom` needs to be in `state`. I think it would be fine to have something like `output_window.viewport_at_bottom` and renderer can just check that.

Yep that's easier to maintain and I could remove some manual reset :smile: thanks for the review!

sudo-tee:
@guillaumeboehm 
This looks pretty reasonable to me.

One minor change. I would ask.

There is a lot of uneeded comments in the MR.

I usually try to put comments on things that are not obious.

lua/opencode/ui/output_window.lua
```diff
@@ -27,6 +28,37 @@ function M.mounted(windows)
   return windows and windows.output_buf and windows.output_win and vim.api.nvim_win_is_valid(windows.output_win)
 end
 
+---Check if the output window is currently at the bottom
+---@param win? integer Window ID, defaults to state.windows.output_win
+---@return boolean true if at bottom, false otherwise
+function M.is_at_bottom(win)
+  -- If always_scroll_to_bottom is enabled, always return true
```
sudo-tee: No need the if is self explanatory

lua/opencode/ui/output_window.lua
```diff
@@ -27,6 +28,37 @@ function M.mounted(windows)
...
+  if config.ui.output.always_scroll_to_bottom then
+    return true
+  end
+
+  win = win or (state.windows and state.windows.output_win)
+
+  if not win or not vim.api.nvim_win_is_valid(win) then
+    return true  -- Assume at bottom if window invalid
```
sudo-tee: Same here it's pretty clear what it does

lua/opencode/ui/output_window.lua
```diff
@@ -27,6 +28,37 @@ function M.mounted(windows)
...
+  end
+
+  if not state.windows or not state.windows.output_buf then
+    return true
+  end
+
+  local ok, line_count = pcall(vim.api.nvim_buf_line_count, state.windows.output_buf)
+  if not ok or not line_count or line_count == 0 then
+    return true  -- Empty buffer, consider at bottom
```
sudo-tee: here also

lua/opencode/ui/output_window.lua
```diff
@@ -177,13 +209,24 @@ function M.setup_autocmds(windows, group)
   state.subscribe('current_permission', function()
     require('opencode.keymap').toggle_permission_keymap(windows.output_buf)
   end)
+
+  -- Track scroll position when window is scrolled
+  vim.api.nvim_create_autocmd('WinScrolled', {
+    group = group,
+    buffer = windows.output_buf,
+    callback = function()
+      -- Update state to track if user is at bottom
```
sudo-tee: Same here, no need for the comment

sudo-tee:
Thanks for fixing my nitpicking :)

PR #106, opened 2025-11-05
fix(renderer): clear permissions when resetting

### LINKED ISSUE #102: bug: The permission dialog does not reapear after reopening the opencode window
Tested on aad6305

When toggling off the opencode window while a permission dialog is shown, it does not reappear again and ends up in a weird hanging state.

I'm testing this by sending a webfetch request to the LLM and closing the window after the permission prompt is shown.

Image

*Closing and reopening*

Image

And it keeps hanging even after sending messages

Image

cameronr:
There currently isn't a way to fetch a pending permission so when we're resetting, reject any pending permissions.

Fixes #102

### DIFF
--- a/lua/opencode/ui/renderer.lua
+++ b/lua/opencode/ui/renderer.lua
@@ -46,7 +46,7 @@ function M.reset()
   state.messages = {}
   state.last_user_message = nil
 
-  if state.current_permission then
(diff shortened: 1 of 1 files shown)


sudo-tee:
Dang you are fast.

PR #107, opened 2025-11-06
fix(core): create session if no current and no last 

cameronr:
If we're not starting a new session, we don't have a current session and
we don't have a last session, we have to create a new session.

Without a session, submitting prompts fails silently.

Clean up the promise chains in server_job (they didn't really make sense since we don't return them and we don't need them anyway)

Also more type / diagnostic cleanup

### DIFF
--- a/lua/opencode/core.lua
+++ b/lua/opencode/core.lua
@@ -76,6 +76,9 @@ function M.open(opts)
   else
     if not state.active_session then
       state.active_session = session.get_last_workspace_session()
+      if not state.active_session then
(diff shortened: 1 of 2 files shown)


cameronr:
@sudo-tee I think this is now probably urgent. We weren't creating a session if you had no active_session and no previous sessions. We haven't noticed because we all have sessions but a new user (or working in a new project) will fail silently when you submit a response without this fix.

sudo-tee:
Very nice fix.

I did a quick review and it all looks fine

Issue #97, opened 2025-11-04
feat: add option to open a buffer for each touched file


mikaoelitiana:
It would be great if there was an option that would allow to follow all the file edition in live and have the buffer opened as opencode progress.

sudo-tee:
This is a nice idea.

Although I would rather provide a hook mechanism where you could implement anything.

Something like 

hooks={
  file_updated = function....
  panel_toggled ...
  session_loaded ...
  
}

Would that fit your needs?

mikaoelitiana:
> This is a nice idea.
> 
> Although I would rather provide a hook mechanism where you could implement anything.
> 
> Something like
> 
> hooks={ file_updated = function.... panel_toggled ... session_loaded ...
> 
> }
> 
> Would that fit your needs?

Yes I think that will do it

cameronr:
@mikaoelitiana I whipped something up. If you want to test it:

```lua
  {
    -- 'sudo-tee/opencode.nvim',
    'cameronr/opencode.nvim',
    branch = 'feat/hooks',
```

PR #109, opened 2025-11-08
feat(hooks): hooks for edit filed, session loaded

cameronr:
Fixes https://github.com/sudo-tee/opencode.nvim/issues/97


sudo-tee:
This is great,

Thanks for tackling this :)

PR #111, opened 2025-11-09
docs(readme): add demo gif

cameronr:
I thought a demo gif might be nice way to show how the plugin works. Totally fine if it's not your style or you'd rather not use it.


sudo-tee:
This is great :)

Thanks

PR #114, opened 2025-11-11
refactor(renderer): clean up event subscriptions

cameronr:
Noticed that we already had a subscription to `state.active_session` and realized we could clean up the subscribes a bit.


sudo-tee:
Great cleanup

Thanks for this

PR #115, opened 2025-11-11
feat(ui): move model + mode to footer ands move token count topbar

sudo-tee:
This makes the panel closer to opencode tui.

image

Model is hiddent when loading for the spinner

image

### DIFF
--- a/lua/opencode/ui/footer.lua
+++ b/lua/opencode/ui/footer.lua
@@ -7,6 +7,10 @@ local loading_animation = require('opencode.ui.loading_animation')
 
 local M = {}
 
+local function utf8_len(str)
+  return vim.fn.strchars(str)
+end
+
 local function get_mode_highlight()
   local mode = (state.current_mode or ''):lower()
   local highlights = {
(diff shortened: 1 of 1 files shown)


cameronr:
Ooh, I'll take a look today!

cameronr:
I think it's a nice improvement. The only thing that's kind of been bothering me is the lack of spacing between the footer and the last line of the assistant response (it feels particularly off because there's more spacing below the footer because the window separator takes up a whole row). This change makes that conflict more noticeable because now there's always something in left part of the footer (instead of that only being where the thinking message is). It's especially noticeable when there's an error or anything with full height characters:

Screenshot 2025-11-11 at 14 34 54

I seem to always suggest right aligning, but maybe that's also an option here? Here's what it looks like:

Screenshot 2025-11-11 at 14 36 06

There's less likely to be output text on right so the spacing issue doesn't have as much visual impact (at least for me). Opencode TUI does also have the model and mode together on the right...

sudo-tee:
You are right. I have been trying to add padding to the bottom for a while... But couldn't get it to work in a reliable way.. 

I tried phantom lines. scroll off on the buffer.  And many other things.

If you have an idea on how to do it let me know.


The issue having it to the right. Is that it looks off when the cancel message is there. Maybe hiding it would make sense.

cameronr:
Yeah, in my quick right align test, I hide the model when there's a request in process:

Screenshot 2025-11-11 at 15 06 13

Haha, I also tried a number of things that didn't work. The only thing I ever got working well enough was to add an extra newline (that got removed when inserting new parts at the end). I wasn't sure if I liked it better or not. I'll push a branch for you to try out, if you want.

sudo-tee:
Ok I will then align it to the right. It looks fine to me

PR #116, opened 2025-11-12
test(replay): fix full session replay

cameronr:
The issue is a bit subtle: the event notification about the session changing from nil to an id gets wrapped in vim.schedule but that means it gets delivered after we've loaded the full session, which causes us to clear the whole session right after we just loaded it, leaving the output blank.

It's possible that we might want to not wrap state notifications in vim.schedule and expect listeners to handle cases where they can't be in a fast context. The primary benefit of that would be guaranteed ordering of events.

For example, in this case, the event that the session was set would've been processed before we loaded the session so the issue wouldn't exist.

The issue is only coming up now because the replay script used to unsubscribe to `state.active_session` changes but i didn't keep that because it felt like a hack.

For now, I think we can just watch out for any issues in actual usage.

### DIFF
--- a/tests/helpers.lua
+++ b/tests/helpers.lua
@@ -21,12 +21,11 @@ function M.replay_setup()
   end
 
   state.current_mode = 'build' -- default mode for tests
-  state.windows = ui.create_windows()
 
-  -- we use the event manager to dispatch events
+  -- we use the event manager to dispatch events, have to setup before ui.create_windows
   require('opencode.event_manager').setup()
 
-  renderer.setup_subscriptions()
+  state.windows = ui.create_windows()
 
   renderer.reset()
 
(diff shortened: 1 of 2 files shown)


cameronr:
Actually, the more I think about this the more I think it's probably not an issue in practice. The replay harness pattern of setting the session and then manually loading the session directly has the order and effect backwards. Normally, it's setting `state.active_session` that should be triggering the loading.

sudo-tee:
> Actually, the more I think about this the more I think it's probably not an issue in practice. The replay harness pattern of setting the session and then manually loading the session directly has the order and effect backwards. Normally, it's setting `state.active_session` that should be triggering the loading.

You are right about this. It makes more sense this way

cameronr:
Sounds good. This PR should still be merged as it just fixes the full session replay.

PR #117, opened 2025-11-12
fix(renderer): add an extra newline when inserting

cameronr:
Forces a blank line between the output content and the footer.

I'm still not quite sure how I feel about it. Sometimes I like it, sometimes I think, "eh, maybe not" so I'm curious to get your opinion.


sudo-tee:
On my part I like it way better with a padding line,

I think just adding this padded lines is the right approach

PR #119, opened 2025-11-13
feat(zoom): add zoom api

### LINKED ISSUE #118: A handy shortcut to zoom in/out the opencode splits?
Hi,

Will it be possible to add a shortcut to zoom in/out the opencode spllits (both the output and the input splits).

My workflow often shows this pattern:

* I want to talk with the agent in details, about the plan and the changes. But the opencode pane is too small for large information. During this time, I want to zoom in the pane and maximize the pane by occupying about 4/5 width of the window.
* Sometimes during the discussion, I need to fact check some files in the editor to understand and evaluate what the agent wants to do. Now I want to restore the pane back to its init normal size.

What do you think of it?

Thanks! Can't live for a second without this plugin!

cameronr:
`config.ui.zoom_width` sets the percentage to zoom

Fixes #118


sudo-tee:
Thanks for the PR 

It looks good to me

Issue #122, opened 2025-11-16
Fix session picker module import and formatting


nakhan98:
## Summary
- change the session picker to require `opencode.api` instead of the nonexistent `lua.opencode.api`
- ensure `vim.ui.select` receives strings so its formatter no longer raises a concat error

## Environment
- `vim-plug` used for plugin management
- Local checkout via `Plug '~/dev/github/opencode.nvim'` in `init.vim`
- macOS 15.7.2 with Neovim 0.11.5
- Opencode version 1.0.65

## Snippets
```lua
-- lua/opencode/ui/session_picker.lua
-local api = require('lua.opencode.api')
+local api = require('opencode.api')
```

```lua
-- lua/opencode/ui/ui.lua
-local parts = { { session.id } }
+local parts = { session.id }
```

## Error Logs
```
Error executing Lua callback: .../lua/opencode/ui/session_picker.lua:5: module 'lua.opencode.api' not found
  no file './lua/opencode/api.lua' ...
  ?
Error executing Lua callback: .../lua/vim/ui.lua:48: invalid value (table) at index 1 in table for 'concat'
Stack traceback: ...
```

## Testing
- not run (per agent instructions)

cameronr:
First issue fixed in #120, second one in #123

PR #123, opened 2025-11-16
fix(ui): select_session with vim.ui.select

cameronr:
Was using an old format.

Show the title if we have one, otherwise show session.id

Fixes https://github.com/sudo-tee/opencode.nvim/issues/122


sudo-tee:
Thanks for the PR

PR #124, opened 2025-11-18
fix(context-actions): fix off by one context actions

sudo-tee:
There was 2 off by one errors when adding contextual actions

When adding them in snapshots

When refreshing them after a part update

A couple of test needed to be regenerated for it to work

I added the possibility to regenerate a test snapshot by it's name or file-name

### DIFF
--- a/tests/helpers.lua
+++ b/tests/helpers.lua
@@ -27,7 +27,10 @@ function M.replay_setup()
 
   state.windows = ui.create_windows()
 
-  renderer.reset()
+  -- disable fetching session and rendering it (we'll handle it at a lower level)
+  renderer.render_full_session = function()
+    return require('opencode.promise').new():resolve(nil)
+  end
 
   M.mock_time_utils()
 
--- a/tests/manual/renderer_replay.lua
+++ b/tests/manual/renderer_replay.lua
@@ -196,14 +196,12 @@ function M.replay_full_session()
   end
 
   state.active_session = helpers.get_session_from_events(M.events, true)
-  vim.schedule(function()
-    local session_data = helpers.load_session_from_events(M.events)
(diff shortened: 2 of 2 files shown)


sudo-tee:
@cameronr  When you have a chance, can you do a quick check/review on this.

I was not seeing the context action on restore points anymore and I investigated.

This is my attempt at fixing it.

cameronr:
Ah, I see the issue. When I added the extra newline at the end, I didn't account for that for actions. I didn't catch it with the replay tests because they needed to be regenerated for the newline so I missed that the actions were off after that change.

I think there's also some inconsistency on whether actions should be 0 indexed or 1 indexed. according to `:h api-indexing` most nvim functions are line 0 indexed with some notable exceptions, including `nvim_win_get_cursor`. I think we should standardize on api-indexing.

If that all sounds good, I'll fix the root issue and clean up the indexing to be consistent.

Also, the replay tool has a `:ReplaySave` command which is what I've used to update a single output. The idea is that you can save the output once you reviewed it to be correct.

sudo-tee:
Sure If you want to fix the root cause I'm all for it.

It was confusing indeed, and the -1 indexing is also confusing 

thanks for your help

cameronr:
Was finally able to push my changes after GH resolved it's issue.

The main changes are here, with the underlying fix being in `renderer.lua`:

[`4099a8f` (#124)](https://github.com/sudo-tee/opencode.nvim/pull/124/commits/4099a8f5fecaca8dce2b42494bf9f37cac1c1d92)

cameronr:
I guess I spoke too soon, the tests failed because GH is still having issues. I'll rerun the tests

sudo-tee:
Thanks for the fixes :)

PR #125, opened 2025-11-19
Add done_thinking and permission_requested hooks to user config

guillaumeboehm:
I felt like having system notifications on certain opencode events would be really nice, so I implemented those two hooks available in the user config.

I had to add a "user_message_count" to the state to track the user initiated requests rather than all requests otherwise the hook would trigger even if the user didn't send anything.
I'm not sure that the mechanism to determine if a job is user initiated is correct, it works for me but there might be edge cases that I'm not aware of.

### DIFF
--- a/lua/opencode/core.lua
+++ b/lua/opencode/core.lua
@@ -147,16 +147,24 @@ function M.send_message(prompt, opts)
   params.parts = context.format_message(prompt, opts.context)
   M.before_run(opts)
 
-  state.user_message_count = state.user_message_count + 1
+  -- Capture the session ID to ensure we track the message count for the correct session
+  local session_id = state.active_session.id
+  local sent_message_count = vim.deepcopy(state.user_message_count)
+  sent_message_count[session_id] = (sent_message_count[session_id] or 0) + 1
+  state.user_message_count = sent_message_count
+
   state.api_client
-    :create_message(state.active_session.id, params)
+    :create_message(session_id, params)
     :and_then(function(response)
       if not response or not response.info or not response.parts then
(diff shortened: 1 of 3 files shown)


lua/opencode/server_job.lua
```diff
@@ -35,6 +43,9 @@ function M.call_api(url, method, body)
   local function remove_from_requests()
     for i, entry in ipairs(M.requests) do
       if entry == request_entry then
+        if is_user_message() then
+          state.user_message_count = state.user_message_count - 1
```
sudo-tee: I think it would be better to set this state in the core.lua file.

Image

The send message function is only called by the user

sudo-tee:
Can you add some test cases in the `hooks_spec.lua` file for the new hooks ?

Also  can you add them to the README file with a short description / example.  I just realised that the doc was missing for the other hooks and added them to the docs

lua/opencode/server_job.lua
guillaumeboehm: Aah right I see, yeah it simplifies everything xD

guillaumeboehm:
Alright, after thinking that adding the session info as a hook param would be nice to access the session title etc, I realized that the hook would only be called for the currently active session.
But the idea was to be notified for longer sessions, so it's better to be notified even after switching session.
I'll let you see if you think that it's looking good enough and I'll do the same thing for the permission hook.

I had to change the two subscribed function in core into "private-ish" functions in order to use them in the unit tests, I'm not sure if that's fine or not.

sudo-tee:
Thanks for the fixes.

It looks fine to me

PR #126, opened 2025-11-20
feat(image): paste image from the clipboard

sudo-tee:
This PR add the possibility to paste image from the clipboard.

You can paste with `<M-v>` in the input window.  Or with the global keymap `<leader>ov` or with `Oppencode paste_image`

The pasted image will be added as a file attachement similar to the mentions.

### DIFF
--- a/lua/opencode/image_handler.lua
+++ b/lua/opencode/image_handler.lua
@@ -12,11 +12,18 @@ local function is_valid_file(path)
   return vim.fn.getfsize(path) > 0
 end
 
---- Run shell command and return success
---- @param cmd string
+--- Run shell or powershell command and return success
+--- @param cmd string|table
+--- @param opts table?
 --- @return boolean
-local function run_shell_cmd(cmd)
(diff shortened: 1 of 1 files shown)


lua/opencode/image_handler.lua
```diff
@@ -0,0 +1,156 @@
...
+local function is_valid_file(path)
+  return vim.fn.getfsize(path) > 0
+end
+
+--- Run shell command and return success
+--- @param cmd string
+--- @return boolean
+local function run_shell_cmd(cmd)
+  return vim.system({ 'sh', '-c', cmd }):wait().code == 0
+end
```
cameronr: purely stylistic, but you could have `run_shell_cmd` take an optional opts so that `save_base64` and others could use it as well.

sudo-tee:
> Wow, I didn't think this was even possible! Very nice!

Me neither, until I saw a tweet about it for opencode, and proceeded to 'vibe port' their implementation with my own tweaks to fit in the plugin

Issue #133, opened 2025-12-05
UI is slow to open with opencode v1+


cameronr:
It looks like opencode is slower to open in versions v1+. It used to be very fast but now it it has a noticeable pause now (1 second for me):

<img width="1688" height="1120" alt="Image" src="https://github.com/user-attachments/assets/32a742a5-8bc1-4b7e-b9e9-8a79ae769008" />

I'm going to take a look and see if I can improve the experience

sudo-tee:
I also realised that and started a fix.  I will push a branch, maybe we can work together on this.

sudo-tee:
https://github.com/sudo-tee/opencode.nvim/pull/134

There is still some issues like the selection context is not loaded when opening the panel.. but work if the panel is already opened.

Then pr is not waiting on the server requests to open the panel, so the perceived speed is way better

Let me know what you think

cameronr:
Oh great! I'll take a look at your branch.

PR #134, opened 2025-12-05
perf: open the panel without waiting on session

### LINKED ISSUE #137: Slow to open Opencode window when session is really, really long.
I think your synchronously reading the files of a session which blocks the UI when the session gets super long.

You should do the file reads asynchronously so the UI is not blocked.

https://github.com/user-attachments/assets/34719b1a-834a-4ed8-bf16-6d92a3a0d1a0

sudo-tee:
This PR is removing the :wait on the critical paths in favor of a coroutine similar to how promises work in js.

This should resolve #137 and #133

### DIFF
--- a/lua/opencode/api.lua
+++ b/lua/opencode/api.lua
@@ -262,7 +262,7 @@ function M.prev_message()
   require('opencode.ui.navigation').goto_prev_message()
 end
 
-function M.submit_input_prompt()
+M.submit_input_prompt = Promise.async(function()
   if state.display_route then
     -- we're displaying /help or something similar, need to clear that and refresh
     -- the session data before sending the command
@@ -271,7 +271,7 @@ function M.submit_input_prompt()
   end
 
   input_window.handle_submit()
-end
+end)
 
 function M.mention_file()
   local picker = require('opencode.ui.file_picker')
@@ -506,8 +506,8 @@ M.mcp = Promise.async(function()
   ui.render_lines(msg)
 end)
 
-function M.commands_list()
-  local commands = config_file.get_user_commands():wait()
+M.commands_list = Promise.async(function()
+  local commands = config_file.get_user_commands():await()
   if not commands then
     vim.notify('No user commands found. Please check your opencode config file.', vim.log.levels.WARN)
     return
@@ -530,7 +530,7 @@ function M.commands_list()
 
   table.insert(msg, '')
   ui.render_lines(msg)
-end
+end)
 
 M.current_model = Promise.async(function()
   return core.initialize_current_model()
--- a/lua/opencode/ui/input_window.lua
+++ b/lua/opencode/ui/input_window.lua
@@ -76,7 +76,7 @@ function M.handle_submit()
 end
 
 M._execute_slash_command = function(command)
-  local slash_commands = require('opencode.api').get_slash_commands()
+  local slash_commands = require('opencode.api').get_slash_commands():await()
   local key = config.get_key_for_function('input_window', 'slash_commands') or '/'
 
   local cmd = command:sub(2):match('^%s*(.-)%s*$')


sudo-tee:
There is still some issues, like test and context not loading properly, but it's a start

cameronr:
Hmm, the UI panels open faster but then I still get a noticeable blocking pause after the panels appear:

Kapture 2025-12-04 at 20 45 58

I think the underlying issues is the `Promise:wait` in `ensure_server`:

https://github.com/sudo-tee/opencode.nvim/blob/9223b98036baeffb42d213a0ee0c7be0d032f1a1/lua/opencode/server_job.lua#L134-L155

Instead of doing `:wait` there, maybe we can just return the promise? Also, since `ensure_server` sets `state.opencode_server`, we don't need to return it from that function.

sudo-tee:
> Hmm, the UI panels open faster but then I still get a noticeable blocking pause after the panels appear:

> Instead of doing `:wait` there, maybe we can just return the promise? Also, since `ensure_server` sets `state.opencode_server`, we don't need to return it from that function.

I will see what I can do.  But even then until the server is loaded you cannot display most of the UI elements.

sudo-tee:
@cameronr 

I pushed a couple of improvements and removed the :wait()

It improved a little bit, there is no jumping of the cursor, but we still have to wait for the session to be loaded. Which in itself is un-avoidable.


https://github.com/user-attachments/assets/42df232d-bb0b-4c9e-a0e9-f4410ec6fa9a

cameronr:
It's improving but I still get a UI freeze. I think the core problem is that `Promse:wait` uses `vim.wait` which blocks the neovim UI. I think it's fine if we're not able to fill in our UI elements until the opencode server starts but we shouldn't be blocking the neovim UI while we're waiting for that and I think that blocking is what makes it "feel" slow.

I added a `debug.traceback()` print to `Promise:wait` and it's called from these places when opening the panels:

```
              waiting
stack traceback:
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:173: in function <...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:162>
	[C]: in function 'pcall'
	...ev/neovim-dev/opencode.nvim/lua/opencode/config_file.lua:13: in function 'get_opencode_config'
	...ev/neovim-dev/opencode.nvim/lua/opencode/config_file.lua:76: in function 'get_opencode_agents'
	...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:373: in function 'ensure_current_mode'
	...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:84: in function <...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:81>
	[C]: in function 'pcall'
	...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:81: in function <...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:80>
	[C]: in function 'pcall'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:81: in function 'cb'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:43: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
              waiting
stack traceback:
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:173: in function <...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:162>
	[C]: in function 'pcall'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/session.lua:45: in function 'get_all_sessions'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/session.lua:83: in function 'get_all_workspace_sessions'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/session.lua:109: in function 'get_last_workspace_session'
	...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:97: in function <...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:81>
	[C]: in function 'pcall'
	...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:81: in function <...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:80>
	[C]: in function 'pcall'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:81: in function 'cb'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:43: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
              waiting
stack traceback:
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:173: in function <...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:162>
	[C]: in function 'pcall'
	...ev/neovim-dev/opencode.nvim/lua/opencode/config_file.lua:31: in function 'get_opencode_project'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/session.lua:9: in function 'project_id'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/session.lua:27: in function 'get_workspace_session_path'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/session.lua:61: in function <...am/Dev/neovim-dev/opencode.nvim/lua/opencode/session.lua:60>
	vim/shared.lua: in function 'get_all_sessions'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/session.lua:83: in function 'get_all_workspace_sessions'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/session.lua:109: in function 'get_last_workspace_session'
	...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:97: in function <...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:81>
	[C]: in function 'pcall'
	...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:81: in function <...s/cam/Dev/neovim-dev/opencode.nvim/lua/opencode/core.lua:80>
	[C]: in function 'pcall'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:81: in function 'cb'
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:43: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
              waiting
stack traceback:
	...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:173: in function <...am/Dev/neovim-dev/opencode.nvim/lua/opencode/promise.lua:162>
	[C]: in function 'pcall'
	...ev/neovim-dev/opencode.nvim/lua/opencode/config_file.lua:48: in function 'get_opencode_providers'
	...ev/neovim-dev/opencode.nvim/lua/opencode/config_file.lua:61: in function 'get_model_info'
	.../Dev/neovim-dev/opencode.nvim/lua/opencode/ui/topbar.lua:19: in function 'format_token_info'
	.../Dev/neovim-dev/opencode.nvim/lua/opencode/ui/topbar.lua:105: in function <.../Dev/neovim-dev/opencode.nvim/lua/opencode/ui/topbar.lua:93>
```

To fully fix it, we'll have to update those places to not use wait in the common path so the nvim UI is never blocked.

sudo-tee:
Good point, I think the "real" way of doing it would be to have a promise based on coroutines.. but it would require a bigger refactor.  I will check the usages of :wait and see if it's possible to continue without blocking the ui or to manage the promise.

sudo-tee:
@cameronr  I had a look at eliminating the :wait on the critical path.

The isssue I have is that promises are spreading and need to be changed everywhere. I gave up after a couple of hours.

If you want to have a jab at it don't hesitate. I will try to find another solution to avoid calling :wait.

cameronr:
> Good point, I think the "real" way of doing it would be to have a promise based on coroutines.. but it would require a bigger refactor. I will check the usages of :wait and see if it's possible to continue without blocking the ui or to manage the promise.

Yes, I think you're right. I'll think about it a bit and see if I can come up with anything but I suspect it would involve refactoring most of the api to take call backs rather than executing synchronously.

sudo-tee:
@cameronr 

Just so you know I'm in the middle of the process of converting every promise into a coroutine. It will take a moment, but I think it will improve and cleanup a lot of stuff.

cameronr:
oh great! switching to coroutines was what i was going to explore so that sounds good!

sudo-tee:
@cameronr 

I think I'm almost done

If you want to try and review

image

Now the profiler is telling the await method is heavy.. but I'm not sure it's reliable for coroutines.

Theses are not blocking waits.

The panel opening feels faster to me.

Let me know what you think

There is still plenty of manual tests to do as this is a pretty significant change.

cameronr:
Ooh, exciting! I'll take a look today!

cameronr:
I've started looking into this and you really have made great progress!

I did notice that there's still a small delay when opening the panels and I traced it down to `core.opencode_ok`. It seems like the opencode app has a noticeable delay now, even when running with just `--version`:

```bash
❯ time opencode --version
1.0.138
opencode --version  0.26s user 0.03s system 50% cpu 0.577 total
```

Rather than using `vim.system():wait` which, i believe, is blocking (even in a coroutine context), we could do something like:

```lua
-- this could be in Promise or util
local function system_async(cmd, opts)
  local p = Promise.new()

  vim.system(cmd, opts or {}, function(result)
    if result.code == 0 then
      p:resolve(result)
    else
      p:reject(result)
    end
  end)

  return p
end


M.opencode_ok = Promise.async(function()
  if vim.fn.executable('opencode') == 0 then
    vim.notify(
      'opencode command not found - please install and configure opencode before using this plugin',
      vim.log.levels.ERROR
    )
    return false
  end

  vim.notify('alive')
  if not state.opencode_cli_version or state.opencode_cli_version == '' then
    local result = system_async({ 'opencode', '--version' }):await()
    -- can test with something like:
    -- local result = system_async({ 'sleep', '3' }):await()
    local out = (result and result.stdout or ''):gsub('%s+$', '')
    state.opencode_cli_version = out:match('(%d+%%.%d+%%.%d+)') or out
    vim.notify('done: ' .. tostring(state.opencode_cli_version))
  end
  ...
end)
```

Let me know if it's ok to make that change.

I'll also keep looking at the changes.

sudo-tee:
Great catch. Yes I am totally fine with this change 

Can you run 

'time opencode --version' on your system.

For me I don't see any significant delay

But the change should help anyway

cameronr:
I had already included the time results in my previous comment :) Takes about 270ms:

> I've started looking into this and you really have made great progress!
> 
> I did notice that there's still a small delay when opening the panels and I traced it down to `core.opencode_ok`. It seems like the opencode app has a noticeable delay now, even when running with just `--version`:
> 
> ```shell
> ❯ time opencode --version
> 1.0.138
> opencode --version  0.26s user 0.03s system 50% cpu 0.577 total
> ```


I'll make that change.

cameronr:
I added some vim.system sleeps (to both `opencode_ok` and `OpencodeServer:spawn` to make sure it's non-blocking and it looks good:

Kapture 2025-12-09 at 17 41 28

 I'll review the rest of code by tomorrow night.

sudo-tee:
> I had already included the time results in my previous comment :) Takes about 270ms:
> 
> > I've started looking into this and you really have made great progress!
> > 
> > I did notice that there's still a small delay when opening the panels and I traced it down to `core.opencode_ok`. It seems like the opencode app has a noticeable delay now, even when running with just `--version`:
> > 
> > ```shell
> > ❯ time opencode --version
> > 1.0.138
> > opencode --version  0.26s user 0.03s system 50% cpu 0.577 total
> > ```
> 
> 
> I'll make that change.

I was on mobile and didn't see it 😂

lua/opencode/ui/completion/engines/base.lua
sudo-tee, line 1: This file is not used yet, 

It ended up here by error, but I plan to use it in a near feature for the completion bug.

lua/opencode/config_file.lua
```diff
@@ -1,64 +1,71 @@
...
-  end)
-  if not ok then
-    vim.notify('Error fetching Opencode providers: ' .. vim.inspect(result), vim.log.levels.ERROR)
+  return M.providers_promise:catch(function(err)
+    vim.notify('Error fetching Opencode providers: ' .. vim.inspect(err), vim.log.levels.ERROR)
     return nil
-  end
-
-  return result --[[@as OpencodeProvidersResponse|nil]]
+  end)
```
cameronr, line 63: since it now returns a promise, might be cleaner to not have it return nil, but something like `Promise.new():resolve(nil)`. Similar with above.

lua/opencode/api.lua
cameronr: I think `submit_input_prompt` also needs to be marked as async as custom user commands may be async:

https://github.com/sudo-tee/opencode.nvim/blob/c019af68adb0d1c70b200a2b1204330115466385/lua/opencode/ui/input_window.lua#L78-L80

I'l check in a fix after this review

lua/opencode/git_review.lua
```diff
@@ -4,19 +4,19 @@ local snapshot = require('opencode.snapshot')
...
-  if not snapshot_dir then
+  if not M.__snapshot_path then
     vim.notify('No snapshot path for the active session.')
     return nil, nil
   end
-  local args = { 'git', '-C', snapshot_dir }
+  local args = { 'git', '-C', M.__snapshot_path }
   vim.list_extend(args, cmd_args)
   local result = vim.system(args, opts or {}):wait()
   if result and result.code == 0 then
```
cameronr, line 22: Could use Promise.system here if you wanted

lua/opencode/config_file.lua
sudo-tee: Inside a catch block it not necessary, it is already wrapped in a Promise

lua/opencode/git_review.lua
sudo-tee: Yeah I tried , and the way this module is layed out it was to complex.
I think at some point I will need to rewrite this module.

Since it is not a critical path I will leave it as it is for the moment.

PR #135, opened 2025-12-07
fix(session_picker): handle nil session after rename

jinzhongjia:
- Add null check for updated_session result
- Resolve promise with nil if session rename fails
- Prevent index lookup on nil session object


jinzhongjia:
This resolve when cancel rename session, opencode.nvim will report an error

jinzhongjia:
It's just a simple conditional check return.

sudo-tee:
Awesome thanks for the fix

PR #143, opened 2025-12-17
fix(history): error when windows closed

### LINKED ISSUE #141: Telescope history picker got nil report
When  preferred_picker = "telescope", I got this:

``` l
E5108: Lua: ...ore/opt/opencode.nvim/lua/opencode/ui/history_picker.lua:101: attempt to index local 'windows' (a nil value)
stack traceback:
        ...ore/opt/opencode.nvim/lua/opencode/ui/history_picker.lua:101: in function 'callback'
        ...k/core/opt/opencode.nvim/lua/opencode/ui/base_picker.lua:122: in function 'run_replace_or_original'
        ...ack/core/opt/telescope.nvim/lua/telescope/actions/mt.lua:65: in function 'key_func'
        .../pack/core/opt/telescope.nvim/lua/telescope/mappings.lua:293: in function <.../pack/core/opt/telescope.nvim/lua/telescope/mappings.lua:
```
Then I checked the code, and found this:

``` lua
-- /opencode.nvim/lua/opencode/ui/history_picker.lua 

-  94       local windows = state.windows
-  95       if not input_window.mounted(windows) then
-  ...
-  98       ---@cast windows { input_win: integer, input_buf: integer }
-  ...
-  101      require('opencode.ui.mention').restore_mentions(windows.input_buf)

```

so I chanaged the code locally  to this one

``` lua
  return base_picker.pick({
    items = history_items,
    format_fn = format_history_item,
    actions = actions,
    callback = function(selected_item)
      if selected_item and callback then
        callback(selected_item.content or selected_item.text)
      elseif selected_item then
        local input_window = require('opencode.ui.input_window')
        local state = require('opencode.state')
        local windows = state.windows
        if not windows or not input_window.mounted(windows) then
          require('opencode.core').open({ focus_input = true })
        end
        ---@cast windows { input_win: integer, input_buf: integer } | nil

        input_window.set_content(selected_item.content or selected_item.text)
        if windows and windows.input_buf then
          require('opencode.ui.mention').restore_mentions(windows.input_buf)
        end
        input_window.focus_input()
      end
    end,
    title = 'Select History Entry',
    width = config.ui.picker_width or 100,
  })
end
```

cameronr:
Handle the case when the history dialog is open and an item is selected before the input/output windows are created.

Fixes #141

### DIFF
--- a/lua/opencode/ui/history_picker.lua
+++ b/lua/opencode/ui/history_picker.lua
@@ -94,6 +94,7 @@ function M.pick(callback)
         local windows = state.windows
         if not input_window.mounted(windows) then
           require('opencode.core').open({ focus_input = true })
+          windows = state.windows
         end
         ---@cast windows { input_win: integer, input_buf: integer }
 


cameronr:
Even with the async changes, I think it's safe to just set windows right after because any yields would come after the `state.windows` gets set in `core.open`

sudo-tee:
Thanks for the quick fix

PR #146, opened 2025-12-19
feat(ui): add support for 'current' position in UI configuration

### LINKED ISSUE #145: Fullscreen none-split option
Would be great to have the option to use the full active buffer window for for the opencode UI instead of forcing a split. This is especially useful for small screen and laptops. But also increasingly, it's not always necessary to look at the code side-by-side. Perhaps in addition to `right` and `left` we can also add `current`. Of course toggling should work to restore the previous active buffer.

aweis89:
## Summary
- Added `'current'` as a valid position option for UI placement, allowing Opencode to take over the active window.
- Updated position handling in dimensions and window closing logic to support this new layout.
- Disabled `winfixbuf` in `'current'` mode and added an autocommand to automatically close Opencode when switching to a different file, enabling a seamless workflow with file pickers.
- Ensured compatibility with existing configurations.

Resolves #145

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ require('opencode').setup({
     }
   },
   ui = {
-    position = 'right', -- 'right' (default) or 'left'. Position of the UI split
+    position = 'right', -- 'right' (default), 'left' or 'current'. Position of the UI split. 'current' uses the current window for the output.
     input_position = 'bottom', -- 'bottom' (default) or 'top'. Position of the input window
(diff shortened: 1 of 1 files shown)


sudo-tee:
Interesting, I will test this today

sudo-tee:
@aweis89  This looks good.

Can you add the config option to the README ?

I will merge after this

sudo-tee:
Thanks for the PR

PR #149, opened 2025-12-19
fix: improve context clearing in new sessions and file mentions

aweis89:
## Summary
- Clear attachments when starting a new session with `/new` command

## Fixes
1. **Context persisting across new sessions**: The `/new` command now calls `context.unload_attachments()` to clear mentioned files, selections, and linter errors when starting a new session

## Test plan
- [x] Run unit tests
- [x] Manually test `/new` command clears context


sudo-tee:
Thanks for the PR

PR #151, opened 2025-12-21
feat: add code reference picker for navigating file references in AI responses

aweis89:
## Summary

- Add a code reference picker (`gr` keybinding in output window) to navigate to file references mentioned in AI responses
- Enable easy navigation when asking exploratory questions about a codebase
- Full support for Snacks, Telescope, and fzf-lua pickers with file preview and line navigation

## Use Case

When asking the AI exploratory questions about a codebase (e.g., "how does the session management work?", "where is error handling implemented?"), the AI often responds with file path references like `file://lua/opencode/session.lua:42`. 

This feature makes those references navigable - press `gr` to open a picker showing all detected references with file preview, then select one to jump directly to that location.

This is specifically designed for **read-only exploration**, not for navigating to edits (which can be found via git pickers like `:Telescope git_status`).

## Features

- **Reference detection** exclusively parses `file://` URIs:
  - `file://path/to/file.lua` (path only)
  - `file://path/to/file.lua:42` (with line number)
  - `file://path/to/file.lua:42:10` (with line and column)
  - `file://path/to/file.lua:42-45` (with line ranges)

- **Picker-specific features**:
  - **Snacks**: File preview with full range highlighting (lines 42-45)
  - **Telescope**: File preview with full range highlighting (lines 42-45)
  - **fzf-lua**: File preview with line navigation (highlights line 42)
  - **Mini.pick**: Basic list without preview

- **Visual icons** are displayed before recognized `file://` URIs in assistant responses, providing immediate visual confirmation that the path will appear in the reference picker
- **System prompt addition**: Instructs the LLM to use `file://` URI scheme and wrap in backticks for file references
- **Deduplication** across messages (most recent first)
- **Validation** - only shows references to files that exist
- **Per-message caching** - references are parsed when the AI finishes responding and cached for fast picker opening
- **Session persistence** - references are parsed when loading previous sessions
- **Keybinding**: `gr` in output window
- **Slash command**: `/references`
- **Vim command**: `:Opencode references`

### Reference Detection & Parsing
- Regex pattern matches `file://` URIs with optional line numbers, columns, and ranges
- Validates files exist on disk before adding to picker
- Caches parsed references per message for performance

### Picker Integration
All pickers use the unified `base_picker` interface with picker-specific optimizations:

#### Snacks
- Uses `pos` and `end_pos` fields for range highlighting
- Background highlight for the entire range

#### Telescope  
- Uses `lnum` and `lnend` fields for range highlighting
- Requires `vim_buffer_vimgrep` previewer (not the default `cat` previewer)
- `TelescopePreviewLine` highlight group for ranges

#### fzf-lua
- Appends `file:line:col:` format after tab separator
- Uses builtin previewer with automatic entry parsing
- Cursorline highlight on start line (range end not supported)

#### Mini.pick
- Basic list display without preview support

### Highlight Groups
- `OpencodeReference` - Icon color for file references in assistant messages
  - Light theme: `#1976D2` (blue)
  - Dark theme: `#7AA2F7` (lighter blue)

## Testing Checklist

- [x] Snacks picker (with file preview and range highlighting)
- [x] Telescope picker (with file preview and range highlighting)
- [x] fzf-lua picker (with file preview and line navigation)
- [x] mini.pick picker (basic list, no preview)
- [x] No picker fallback (returns false gracefully)
- [x] All tests passing (74/74)
- [x] CI passing on all platforms

### DIFF
--- a/lua/opencode/ui/base_picker.lua
+++ b/lua/opencode/ui/base_picker.lua
@@ -213,7 +213,9 @@ local function fzf_ui(opts)
       previewer = opts.preview == 'file' and 'builtin' or nil,
       fn_fzf_index = function(line)
         -- Strip the appended file:line:col info before matching
-        local display_part = line:match('^([^\t]+)') or line
+        -- fzf-lua uses nbsp (U+2002 EN SPACE) as separator
+        local nbsp = '\xe2\x80\x82'
+        local display_part = line:match('^([^' .. nbsp .. ']+)') or line
         for i, item in ipairs(opts.items) do
           if opts.format_fn(item):to_string() == display_part then
             return i
@@ -232,7 +234,7 @@ local function fzf_ui(opts)
         -- For file preview support, append file:line:col format
         -- fzf-lua's builtin previewer automatically parses this format
         if opts.preview == 'file' and type(item) == 'table' then
-          local file_path = item.file or item.file_path or item.path or item.filename
+          local file_path = item.file_path or item.path or item.filename or item.file
           local line = item.line or item.lnum
           local col = item.column or item.col
           
@@ -246,8 +248,10 @@ local function fzf_ui(opts)
               end
               pos_info = pos_info .. ':'
             end
-            -- Append position info after tab separator (fzf-lua standard)
-            line_str = line_str .. '\t' .. pos_info
+            -- Append position info after nbsp separator (fzf-lua standard)
+            -- nbsp is U+2002 EN SPACE, not regular tab
+            local nbsp = '\xe2\x80\x82'
+            line_str = line_str .. nbsp .. pos_info
           end
         end
         


sudo-tee:
This is pretty neat. I will have a try today or tomorrow. But it seems really useful.

sudo-tee:
I played a little bit with this. 

It seems to work as intended.

I have a couple of small comments:

I think it would be better if we could also trigger with ta keymam in the input window. I know there is the /references but having the leymap is faster.

File previews are supported by fzf-lua and Telescope, I think we should support it.

Can you add the new `/references` and `Opencode references`  to the README

aaronweisberg:
- Added full support for `Telescope` file previews with range highlighting
- Added preview support for fzf-lua as well (however it doesn't have an explicit API for range highlights, so it just highlights the first line when there's a range)

sudo-tee:
Looks great to me

Thanks for the PR

PR #154, opened 2025-12-25
feat(ui): add auto_scroll config for output window

jugarpeupv:
Introduce a new `auto_scroll` option in the output window configuration. This allows users to control whether the output window should automatically scroll to the bottom when new content is rendered. The `always_scroll_to_bottom` option still takes precedence if enabled.

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -124,6 +124,7 @@ M.defaults = {
         show_output = true,
         show_reasoning_output = true,
       },
+      auto_scroll = false,
       always_scroll_to_bottom = false,
     },
     input = {
--- a/lua/opencode/ui/output_window.lua
+++ b/lua/opencode/ui/output_window.lua
@@ -232,4 +232,12 @@ function M.clear()
   M.viewport_at_bottom = true
 end
 
+function M.should_auto_scroll()
+  if config.ui.output.always_scroll_to_bottom then
+    return true
+  end
+
+  return config.ui.output.auto_scroll ~= false
+end
+
 return M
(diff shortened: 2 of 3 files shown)


jugarpeupv:
@sudo-tee ey! thanks for this plugin, when opencode response is rendered and i move focus to that window, if i want to copy one of the last sentences, i cant because there is an auto scroll behaviour by default

Ideally it should only auto scroll down once the llm responds, and not every time, the correct way to implemente it is like it is now, like in a terminal buffer, where if the cursor is at the end, the new content is appeded, and the cursor keeps at the end. However if the cursor is not at the end it means the user is reviewing something so we should not auto scroll

This behaviour was a little bit more involved, so till we get to it i simply coded a configuration option to disable auto_scroll entierly which makes it at least usable if you want to copy the last part of a response

sudo-tee:
Thanks for the PR, 

I looked at the code and I fail to see why setting the `always_scroll_to_bottom`  would not work ?

Maybe I am  missing something.

There might be something to fix in the autoscroll behavior though

jugarpeupv:
@sudo-tee you can see the problem in this video, if i want to visually select something from the output pane which is in the last lines i can't because there is an autoscroll behaviour by default

Ideally the autoscroll to bottom should only be performed once the llmm responds and in case the cursor is near the end

https://github.com/user-attachments/assets/28750d94-6260-4c2d-b4dd-95a852125abf

sudo-tee:
Ok I see the issue, Does it happen every time ?

This is rather strange as I often copy and paste from the pane and not experiencing this.

I will have a look to see if I can reproduce it.

sudo-tee:
I will merge it, but will have a look at a better solution.

Thanks for the PR :)

sudo-tee:
@jugarpeupv 
Unfortunately I have to revert this commit.

Now the output never scrolls when opening an old session. Which leads to missing some permissions request, and never seeing new content.

PR #156, opened 2025-12-27
feat(ui): add dynamic input window toggle

aweis89:
## Summary

- Add ability to toggle (hide/show) the input window independently of the output window
- Add new `ui.input.dynamic` config option that automatically hides the input window when focus moves to the output window or when a prompt is submitted

## Motivation

When working with AI responses, the output window benefits from maximum screen real estate. The input window (especially when configured larger) takes up space that isn't needed while waiting for or reading responses. This feature allows the input window to stay out of the way until you're ready to type the next prompt.

## Features

**Toggle Input Window**
- New `toggle_input` API function and command (`:Opencode toggle_input`)
- Default keymap: `<M-i>` (Alt+i) in both input and output windows
- When hidden, the input window is completely closed (not just minimized) to give full space to the output
- Input buffer content is preserved when hidden

**Dynamic Input Mode** (`ui.input.auto_hide = true`)
- When enabled, the input window automatically hides when:
  - Submitting a prompt (`<cr>`)
  - Focus moves to the output window (via any method - keymaps, mouse, `<C-w>` commands, etc.)
- The input window automatically shows when focusing it via `focus_input`, `open_input`, or `toggle_pane`
- When toggling the UI with dynamic mode enabled, no longer defaults to last focused window (always focuses input)

### Auto-scroll preservation during dynamic input resize
When the input window hides/shows with `dynamic = true`, the output window resizes. This could incorrectly update `viewport_at_bottom` via the `WinScrolled` autocmd, breaking auto-scroll. Fixed by:
- Capturing `was_at_bottom` state before hide/show
- Scheduling `scroll_to_bottom()` after resize if previously at bottom

### Tab keymap conflicts with completion plugins
- **Fix:** Restricted the default `<Tab>` toggle focus mapping to only trigger in normal mode, preventing conflicts with auto-complete mappings that use `<Tab>` in insert mode

### Date formatting tests fail on year boundaries (CI)
Tests for `util.format_time` failed when run on year boundaries (e.g., Jan 1, 2026):
- "formats yesterday" expected `31 Dec 09:46 PM` but got `31 Dec 2025 09:46 PM` (year included because Dec 31 is a different year than Jan 1)
- "formats last week" had the same issue
- "handles large millisecond timestamps" also failed due to year comparison

**Fix:** Made tests year-boundary-aware by checking if test dates fall in the same year as today before asserting the expected format pattern.

## Files Changed

- `lua/opencode/api.lua` - Add `toggle_input`, update `submit_input_prompt`, `toggle`
- `lua/opencode/config.lua` - Add `ui.input.dynamic` option and keymaps
- `lua/opencode/ui/autocmds.lua` - Prevent closing all windows during toggle
- `lua/opencode/ui/input_window.lua` - Add `toggle`, `_hide`, `_show`, `is_hidden` functions; preserve scroll position
- `lua/opencode/ui/output_window.lua` - Auto-hide input on focus when dynamic enabled
- `lua/opencode/ui/ui.lua` - Handle hidden state in `focus_input`, simplify `toggle_pane`
- `tests/unit/util_spec.lua` - Fix year-boundary-aware date formatting tests

### DIFF
--- a/lua/opencode/api.lua
+++ b/lua/opencode/api.lua
@@ -1025,9 +1025,9 @@ M.commands = {
   quick_chat = {
     desc = 'Quick chat with current buffer or visual selection',
     fn = M.quick_chat,
-    range = true,
-    nargs = '+',
-    complete = false,
+    range = true, -- Enable range support for visual selections
+    nargs = '+', -- Allow multiple arguments
+    complete = false, -- No completion for custom messages
   },
 
   swap = {


aweis89:
@sudo-tee curious to hear if you had any thoughts on this if you had a chance to try it out.
- Perhaps `dynamic = true` should not be the default?
- It is more inline with the way the TUI works, where the input window is hiddent during model output I believe
- The main advantage for me is having a larger input window without having to compromise on output window size
- Hitting `i` from the output window, to get the input window (in insert mode) has been perferable to a static input window in my experience but curious if you have reservations

sudo-tee:
I didn't had time to look at it yet. Been busy with the holidays and family. But I will look at it pretty soon

lua/opencode/config.lua
```diff
@@ -130,6 +132,8 @@ M.defaults = {
       text = {
         wrap = false,
       },
+      -- Auto-hide input window when prompt is submitted or focus switches to output window
+      dynamic = true,
```
sudo-tee: I would not put this as the default to true.

lua/opencode/ui/ui.lua
```diff
@@ -124,7 +124,17 @@ end
 function M.focus_input(opts)
   opts = opts or {}
   local windows = state.windows
-  if not windows or not windows.input_win then
+  if not windows then
+    return
+  end
+
+  -- Show input window if it's hidden
```
sudo-tee: There is no need for comments where it's obvious

lua/opencode/config.lua
sudo-tee: I think the config should be named auto_hide dynamic is a bit to generic.

sudo-tee (commented):
I like the idea. 

However the implementation feels a bit rough. There is a good amount of layout shifting when toggling the input.

When I move the cursor to the code window the input should also auto_hide, as it feels weird that it hide when focusing back to the opencode window. I attached a video

https://github.com/user-attachments/assets/3c74fb71-5def-4fc3-af4c-df160a304ed2

I think a better solution would be to make the input a floating window when in auto_hide mode, it would reduce the layout shifting. 

What do you think ?

aaronweisberg:
> I like the idea.
> 
> However the implementation feels a bit rough. There is a good amount of layout shifting when toggling the input.
> 
> When I move the cursor to the code window the input should also auto_hide, as it feels weird that it hide when focusing back to the opencode window. I attached a video
> Screen.Recording.2026-01-05.074018.mp4

I mostly used this in full-screen mode `position = "current"` so I didn't consider the intended behavior when switching to another split. I think the input window should be hidden whenever the focus shifts away from the input window instead. I think this will feel less weird. Curious to hear what you think if you're able to try again though.

> I think a better solution would be to make the input a floating window when in auto_hide mode, it would reduce the layout shifting.
> 
> What do you think ?

I think a floating window would add some additional complexities. Ideally when the input window is being hidden, we want to the output window to maximize it's size (so static padding wouldn't be ideal), we also don't want to hide the latest text of the output window behind the input window when the input window is open. Perhaps this can be explored in a future dedicated PR if there's a way to make this better through a float though I'm a bit skeptical.

sudo-tee:
> 
> I think a floating window would add some additional complexities. Ideally when the input window is being hidden, we want to the output window to maximize it's size (so static padding wouldn't be ideal), we also don't want to hide the latest text of the output window behind the input window when the input window is open. Perhaps this can be explored in a future dedicated PR if there's a way to make this better through a float though I'm a bit skeptical.

This is reasonable as long as it's not  the default

lua/opencode/ui/input_window.lua
```diff
@@ -365,4 +390,103 @@ function M.setup_autocmds(windows, group)
...
+---When hidden, the input window is closed entirely
+---When shown, the input window is recreated
+function M.toggle()
+  local windows = state.windows
+  if not windows then
+    return
+  end
+
+  if M._hidden then
+    -- Show: recreate the input window
```
sudo-tee: We don't need comments like this when the code is self explanatory.

lua/opencode/ui/input_window.lua
```diff
@@ -365,4 +390,103 @@ function M.setup_autocmds(windows, group)
...
+  local was_at_bottom = output_window.viewport_at_bottom
+
+  M._hidden = true
+  M._toggling = true
+
+  -- Close the input window (but keep the buffer)
+  pcall(vim.api.nvim_win_close, windows.input_win, false)
+  windows.input_win = nil
+
+  -- Reset toggling flag after the WinClosed event has been processed
```
sudo-tee: please remove comment

lua/opencode/ui/input_window.lua
```diff
@@ -365,4 +390,103 @@ function M.setup_autocmds(windows, group)
...
+  vim.schedule(function()
+    M._toggling = false
+  end)
+
+  -- Focus output window
```
sudo-tee: please remove comment

lua/opencode/ui/input_window.lua
```diff
@@ -365,4 +390,103 @@ function M.setup_autocmds(windows, group)
...
+  -- Don't recreate if already visible
+  if windows.input_win and vim.api.nvim_win_is_valid(windows.input_win) then
+    M._hidden = false
+    return
+  end
+
+  local output_window = require('opencode.ui.output_window')
+  local was_at_bottom = output_window.viewport_at_bottom
+
+  -- Create a new split for the input window
```
sudo-tee: please remove comment

lua/opencode/ui/input_window.lua
```diff
@@ -365,4 +390,103 @@ function M.setup_autocmds(windows, group)
...
+  local output_win = windows.output_win
+  vim.api.nvim_set_current_win(output_win)
+
+  local input_position = config.ui.input_position or 'bottom'
+  vim.cmd((input_position == 'top' and 'aboveleft' or 'belowright') .. ' split')
+  local input_win = vim.api.nvim_get_current_win()
+
+  -- Set the buffer
```
sudo-tee: please remove comment

sudo-tee (approved):
Looks fine now as it is not the default behavior.

The only thing needed is to remove unnecessary comments. Most of the time it's pretty obvious on what the code does. I commented on a couple of ones as example.

aweis89:
Fixed some auto-scrolling so the dynamic input window doesn't block the latest text.
I think this should be good to go 🎉

sudo-tee:
Thanks for the fixes.

It should be good to go

PR #157, opened 2025-12-29
fix: global keymaps config is not respected

jugarpeupv:
It seems to me that global_keymaps config value is not respected, you can see in the video once i open opencode and move the focus to other pane outside of it and press a, then there is a warning from opencode

"no permission request to accept"

https://github.com/user-attachments/assets/dd8651de-5c32-4290-a012-8fa0cfded59e

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -248,11 +248,11 @@ end
 function M.setup(opts)
   opts = opts or {}
 
+  M.values = vim.tbl_deep_extend('force', M.values, opts --[[@as OpencodeConfig]])
+
   if opts.default_global_keymaps == false then
     M.values.keymap.editor = {}
   end
-
-  M.values = vim.tbl_deep_extend('force', M.values, opts --[[@as OpencodeConfig]])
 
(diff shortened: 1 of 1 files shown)


sudo-tee:
Thanks for the fix.

l00sed:
Hey @sudo-tee, for some reason this change broke my keymaps (configured under `M.defaults.keymap.editor`). I reverted this change locally and they started working again.

I think there might be an issue with the provided fix...

```lua
-- Default configuration
---@type OpencodeConfig
M.defaults = {
  preferred_picker = nil,
  preferred_completion = nil,
  default_global_keymaps = true,
  default_mode = 'build',
  legacy_commands = true,
  keymap_prefix = '<leader>o',
  keymap = {
    editor = {
      ['<leader>og'] = { 'toggle', desc = 'Toggle Opencode window' },
      ['<leader>oi'] = { 'open_input', desc = 'Open input window' },
      ['<leader>oI'] = { 'open_input_new_session', desc = 'Open input (new session)' },
```

sudo-tee:
Thanks. I will have a look after the holidays.

PR #159, opened 2026-01-02
feat(server): share process across multiple instances

jinzhongjia:
Description:

  Problem

  The current file lock mechanism for preventing multiple opencode server instances is unstable:
  - Multiple opencode processes can be created due to TOCTOU race conditions
  - Server process not properly cleaned up when neovim exits

  Solution

  Replace the dual-file lock mechanism with a simpler, more reliable approach:

  - Fixed port binding (default: 41096) as atomic mutex - port binding is an OS-level atomic operation
  - Single lock file - removed .flock file, simplified to single file for client tracking only
  - Blocking shutdown - VimLeavePre now waits for cleanup to complete
  - Graceful error handling - clear error message when port is occupied
  - Cross-platform support - Windows taskkill, Unix SIGTERM/SIGKILL
  - User configurable port - config.server.port

  Changes

  - `lua/opencode/opencode_server.lua` - Remove flock, blocking shutdown, fixed port spawn
  - `lua/opencode/server_job.lua` - Health check first, port binding as mutex
  - `lua/opencode/config.lua` - Add server.port config
  - `tests/unit/server_job_spec.lua` - Update tests

### DIFF
--- a/lua/opencode/opencode_server.lua
+++ b/lua/opencode/opencode_server.lua
@@ -147,11 +147,15 @@ local function register_client()
   return true
 end
 
---- @return number|nil server_pid_to_kill
+--- @class UnregisterResult
+--- @field server_pid number|nil
+--- @field is_last_client boolean
+
+--- @return UnregisterResult
 local function unregister_client()
   local data = read_lock_file()
   if not data then
-    return nil
+    return { server_pid = nil, is_last_client = false }
   end
 
   data = cleanup_dead_pids(data)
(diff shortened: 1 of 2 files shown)


jinzhongjia:
This PR is generated by AI with `opencode` and `oh my opencode`, but I have reviewed this code carefully.

I think this code is ok

sudo-tee:
Thanks for the PR this looks interesting, I will play with it a little bit

jinzhongjia:
If you need to modify code, feel free to contact me!

jinzhongjia:
This repository is great

lua/opencode/opencode_server.lua
```diff
@@ -28,27 +242,104 @@ function OpencodeServer.new()
...
+
+  if not server_url then
+    return nil
+  end
+
+  local curl = require('opencode.curl')
+  local result_received = false
+  local is_healthy = false
+
+  curl.request({
```
sudo-tee: It would be cleaner to use the `api-client.lua` and add this route to it.

The api-client is already creating promises for the curl calls, so you can avoid using the `vim.wait` there

lua/opencode/opencode_server.lua
```diff
@@ -28,27 +242,104 @@ function OpencodeServer.new()
...
+end
+
+--- @param url string
+--- @return OpencodeServer
+function OpencodeServer.from_existing(url)
+  local server = OpencodeServer.new()
+  server.url = url
+  server.is_owner = false
+  register_client()
+  server.spawn_promise:resolve(server --[[@as any]])
```
sudo-tee: should cast it to `OpencodeServer`

sudo-tee (commented):
This is a really great feature for the plugin, and it looks stable from the tests I did.

I did add some little nitpicking , but otherwise this looks good

aweis89:
Seems like this will also have the benefit of correctly attaching new sessions to the CWD if that changes within the Neovim instance (e.g. using the `:cd ..` command)

jinzhongjia:
I have modified the code based on the comment, I think now maybe it is ok?

jinzhongjia:
> Seems like this will also have the benefit of correctly attaching new sessions to the CWD if that changes within the Neovim instance (e.g. using the `:cd ..` command)

Theoretically yes, but I'm wondering if anyone would actually change the CWD with `cd` command?
Maybe the system management? But as I know, they all use pure vim or neovim without third plugins

jinzhongjia:
Oh, I found a problem, sometime plugin will create more than one opencode, I will figure out the reason

jinzhongjia:
Shit, I will create a new pr to achieve this

PR #162, opened 2026-01-05
fix(permissions): fix permission request with sub-agents

### LINKED ISSUE #158: No updates and interactive permissions requests when sub-agents triggers tasks
When calling @some-sub-agent configured for opencode, there are no interactive permission requests, and the notification is not always visible or triggered. So the behavior looks like infinite thinking without any actions.

And, when some task is in progress, there are no updates from it, what it does, why, etc. It shows only the task name, asks for some permissions in the background, which is not good that the AI does something in the background, I think

Image

Image

sudo-tee:
The permission request seems to have changed it's format

I updated the permission handling to support both format

This should fix #158

### DIFF
--- a/lua/opencode/ui/formatter.lua
+++ b/lua/opencode/ui/formatter.lua
@@ -690,7 +690,11 @@ function M._format_tool(output, part)
         and state.current_permission.tool.messageID == part.messageID
       )
       ---@TODO this is for backward compatibility, remove later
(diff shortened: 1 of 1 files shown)


romankovalov:
The issue seems to be present after the fix - the only temporary notification shown about permissions
I'm on `f128f20`

image

PR #167, opened 2026-01-07
fix: add option to specify picker layout for snacks picker used in Opencode

MrBanh:
The default snacks picker layout is used for various sources that is not always ideal for this plugin's use case. E.g., many layouts have preview = 'file', which is useless for the session_picker. There's no way to disable preview just for Opencode's snacks picker without affecting other uses of snacks picker for other sources. Looks like it's because of [this change](https://github.com/sudo-tee/opencode.nvim/pull/155)

For example,

```lua
{
  "folke/snacks.nvim",
  opts = {
    picker = {
      layout = { preset = "ivy" }
    }
  }
}
```

The `ivy` preset has a preview window that doesn't work with session_picker, but I don't want to disable it because the default picker layout is used for a bunch of [builtin sources](https://github.com/folke/snacks.nvim/blob/main/docs/picker.md#-sources) where I rely on the file preview.

Screenshot from 2026-01-07 04-24-49

This change allows users to specify a layout in the config. I only added the option for snacks, but we can extend it to the other pickers as well.

There's 3 main ways on how to change the snacks picker layout
1. Don't specify the new options -> it'll just default to the user's snack picker layout preset
2. Specify the new options for opencode, e.g.

    ```lua
    require("opencode").setup({
      ui = {
        picker = {
          snacks_layout = {
            layout = { border = "none", box = "vertical", ... }
          },
        },
      },
    })
    ```

3. Specify a preset available for snacks picker OR a custom layout

    ```lua
    -- opencode.lua
    require("opencode").setup({
      ui = {
        picker = {
          snacks_layout = {
            preset = "custom_layout" -- or builtin snacks, like "select", "default", etc
          },
        },
      },
    })    
    ```

    ```lua
    -- snacks.lua
    { 
      "folke/snacks.nvim",
      opts = {
        picker = {  
          layouts = {  
            custom_layout  = {
              layout = { border = "none", box = "vertical", ... }
              -- ...
    }
    ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
   <img src="https://raw.githubusercontent.com/sst/opencode/dev/packages/web/src/assets/logo-ornate-dark.svg" alt="Opencode logo" width="30%" />
 </div>
 
-### Quick buffer chat (<leader>o/) EXPERIMENTAL:
+### Quick buffer chat (<leader>o/) EXPERIMENTAL
 
 This is an experimental feature that allows you to chat with the AI using the current buffer context. In visual mode, it captures the selected text as context, while in normal mode, it uses the current line. The AI will respond with quick edits to the files that are applied by the plugin.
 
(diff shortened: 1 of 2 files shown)


sudo-tee:
This is a good idea, the implementation seem reasonable also

Thanks for the PR

PR #169, opened 2026-01-08
feat(keymap): allow disabling specific keymaps by setting to false

epheien:
## Summary
- Add support for disabling specific default keymaps by setting them to `false` in configuration
- Users can now selectively disable keymaps they don't need while keeping others

## Changes
- Modified `lua/opencode/keymap.lua`: Added check for `false` values in `process_keymap_entry`
- Updated `README.md`: Added "Disabling Specific Keymaps" section with examples

## Example
```lua
require('opencode').setup({
 keymap = {
   input_window = {
     ['<cr>'] = false,          -- Disable Enter key for submitting prompts
     ['<esc>'] = false,         -- Disable ESC key for closing
     -- Other keymaps not specified will keep their default bindings
   }
 }
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,21 @@ Each keymap entry is a table consising of:
 - An optional mode: `{ 'toggle', mode = { 'n', 'i' } }`
 - An optional desc: `{'toggle', desc = 'Toggle Opencode' }`
 
+#### Disabling Specific Keymaps
+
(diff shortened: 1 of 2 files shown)


sudo-tee:
Thanks for the PR.

The functionality is already there.

But since  you added a nice documentation , plus you made it more explicit in the code, it is fine by me.

PR #174, opened 2026-01-10
fix(ui): use picker.select instead of vim.ui.select to fix snacks picker bug

aweis89:
## Summary

This PR fixes an issue with Snacks.nvim picker when using `vim.ui.select`. The bug causes an "Invalid 'height': Number is not integral" depending on the picker's config.

**Related issue:** https://github.com/folke/snacks.nvim/issues/2539

## Changes

- Added a `picker.select()` function that uses Snacks.picker directly when Snacks is the preferred picker, bypassing the buggy `vim.ui.select` implementation
- For all other pickers (Telescope, FZF-Lua, mini.pick), falls back to `vim.ui.select` to respect user customizations (e.g., dressing.nvim)
- Replaced `vim.ui.select` calls with `picker.select()` in files that use selection UI

## Files Changed

- `lua/opencode/ui/picker.lua` - Added `select()` and `_snacks_select()` functions
- `lua/opencode/api.lua` - Use `picker.select` for agent selection
- `lua/opencode/provider.lua` - Use `picker.select` for model selection  
- `lua/opencode/git_review.lua` - Use `picker.select` for file selection
- `lua/opencode/ui/input_window.lua` - Use `picker.select` for confirmation prompts
- `lua/opencode/ui/ui.lua` - Use `picker.select` for session selection

### DIFF
--- a/lua/opencode/api.lua
+++ b/lua/opencode/api.lua
@@ -422,7 +422,8 @@ end
 
 M.select_agent = Promise.async(function()
   local modes = config_file.get_opencode_agents():await()
-  vim.ui.select(modes, {
+  local picker = require('opencode.ui.picker')
+  picker.select(modes, {
     prompt = 'Select mode:',
   }, function(selection)
     if not selection then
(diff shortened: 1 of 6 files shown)


sudo-tee:
Thanks for the PR. 

I will have a look this weekend

PR #175, opened 2026-01-10
fix(ui): restore window settings when using position=current

aweis89:
## Summary

When using `position = 'current'` configuration, opencode modifies window options (wrap, number, relativenumber, signcolumn, etc.) on the current window. This PR ensures those settings are saved before modification and restored when opencode is closed.

## Problem

Previously, when closing opencode with `position = 'current'`, the window would retain opencode's settings (e.g., `wrap=true`, `number=false`) instead of reverting to the user's original settings.

## Solution

- Save original window option values before modifying them (only when `position = 'current'`)
- Restore saved values when closing opencode
- Store saved options in `state.saved_window_options`

## Files Changed

- `lua/opencode/state.lua` - Added `saved_window_options` field to state
- `lua/opencode/ui/output_window.lua` - Helper function `set_win_option()` that saves original values before setting
- `lua/opencode/ui/ui.lua` - Restore saved options in `close_windows()`

### DIFF
--- a/lua/opencode/state.lua
+++ b/lua/opencode/state.lua
@@ -16,6 +16,7 @@
 ---@field last_output_window_position integer[]|nil
 ---@field last_code_win_before_opencode integer|nil
 ---@field current_code_buf number|nil
+---@field saved_window_options table|nil
 ---@field display_route any|nil
 ---@field current_mode string
 ---@field last_output number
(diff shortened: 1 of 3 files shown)


sudo-tee:
Thanks for the fix

PR #176, opened 2026-01-11
fix(ui): improve task tool output formatting

aweis89:
## Summary
- Display agent type in task description (e.g., `@explore`)
- Show task summary items with status icons (completed/running/pending/error)
- Strip task_metadata tags from output for cleaner display
- Fix metadata.summary structure handling (items have `{id, tool, state}` not full parts)
- Add TaskToolSummaryItem type definition

## Problem
The task tool output was not displaying correctly because the code expected `metadata.summary` to contain full message parts but it actually contains a simpler structure with `{id, tool, state}` causing the UI to hang.

## Solution
Updated the formatter to properly handle the actual summary structure and display tool names with their titles for better context.

### DIFF
--- a/lua/opencode/types.lua
+++ b/lua/opencode/types.lua
@@ -219,7 +219,8 @@
 ---@field message string|nil Optional status or error message
 
 ---@class TaskToolMetadata: ToolMetadataBase
----@field summary OpencodeMessagePart[]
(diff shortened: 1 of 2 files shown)


sudo-tee:
Thanks for the PR.

Look great to me

PR #179, opened 2026-01-12
docs: add Snacks picker integration for opencode.nvim

aweis89:
## Summary
- Add documentation showing how to configure a custom action in all Snacks file pickers (files, git_files, buffers, git_status, etc.) to send selected files to opencode.nvim context

## Test plan
- [x] Verify the code example works with Snacks picker
- [x] Confirm `<localleader>o` sends selected files to opencode context

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,9 +415,8 @@ You can configure a custom action in Snacks pickers to send selected files direc
             end
             picker:close()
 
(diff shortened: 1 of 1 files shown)


sudo-tee:
Very great idea :)

Thanks for the docs

PR #180, opened 2026-01-13
feat: allow using alternative `opencode` binaries

PowerUser64:
This PR adds support for using alternative `opencode` binaries. This is useful if you want to use an opencode binary that isn't in your PATH or if you want to use a wrapper script to run opencode. Tests pass on my machine with `./run_tests.sh`.

(this change was done by hand, then I had an LLM double-check I found all uses of the binary name)


sudo-tee:
Thanks for the PR

this is a nice feature to have, 

It works fine on my end

PR #184, opened 2026-01-14
fix: defer input window keymaps to blink.cmp when visible

### LINKED ISSUE #183: Input window keymaps should defer to blink.cmp when completion menu is visible
## Problem

When blink.cmp autocomplete is enabled, input window keymaps take precedence over blink's mappings even when the completion menu is visible. This is because buffer-local keymaps (set via `vim.keymap.set` with `buffer` option) have higher priority than plugin mappings. This means you cannot use the same mappings for completion and the input window or else it won't work for completion.

Ideally when completion menu is open, the completion mappings should take precedence. Input mappings should essentially "pass-through" to completion when the completion menu is open.

aweis89:
## Summary
- When blink.cmp completion menu is visible, input window keymaps now defer to blink instead of taking precedence
- All input window keymaps are wrapped with visibility checks
- Fixes #183


sudo-tee:
This seems reasonable. 

Thanks for the PR

PR #196, opened 2026-01-22
refactor(code-references): remove reliance on system prompt

### LINKED ISSUE #194: Confusing system prompt
[There is system prompt](https://github.com/sudo-tee/opencode.nvim/blob/main/lua/opencode/core.lua#L176) that instruct LLM to write references using `file://` URI scheme, but [opencode has system prompt](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/session/prompt/codex_header.txt#L77) for codex that says exactly opposite. Because of this, OpenAI models spend half their reasoning budget thinking about how to write file references correctly.

sudo-tee:
- update the reference parser to be more generic instead of relying on a system prompt
- add edit/read files to the references

This should fix #194

### DIFF
--- a/lua/opencode/ui/reference_picker.lua
+++ b/lua/opencode/ui/reference_picker.lua
@@ -18,8 +18,7 @@ local function is_valid_file_reference(path, context)
     return false
   end
 
-  -- Must have extension and exist
-  return path:match('%.[%w]+$') and vim.fn.filereadable(path) == 1
+  return (path:match('%.[%w]+$') and vim.fn.filereadable(path) == 1) or false
 end
 
 ---@class CodeReference
@@ -105,7 +104,7 @@ local function parse_references_with_pattern(text, pattern, message_id)
     local context_start = math.max(1, match_start - 30)
     local context = text:sub(context_start, match_end + 10)
 
-    if is_valid_file_reference(path, context) then
+    if path and is_valid_file_reference(path, context) then
       local line, column, end_line = parse_position_info(line_str, col_or_end_str, end_line_str)
       local ref = create_code_reference(path, line, column, end_line, message_id, match_start, match_end)
       table.insert(references, ref)
@@ -177,8 +176,8 @@ function M.collect_references()
   local seen = {}
   local deduplicated = {}
   for _, ref in ipairs(all_references) do
-    local relative_path = vim.fn.fnamemodify(ref.file_path, ':~:.')
-    local dedup_key = relative_path .. ':' .. (ref.line or 0)
+    local normalized_path = vim.fn.fnamemodify(ref.file_path, ':p')
+    local dedup_key = normalized_path .. ':' .. (ref.line or 0)
     if not seen[dedup_key] then
       seen[dedup_key] = true
       table.insert(deduplicated, ref)


sudo-tee:
@aweis89 I refactored the references collection to not rely on a system prompt, and prefer parsing common file patterns.

Let me know if this works as you expect.

aweis89:
## False Positive Concerns

Thanks for this refactor! Removing the system prompt dependency improves provider compatibility. However, I have concerns about false positives with the more permissive patterns:

### Potential False Positives

**1. Plain path pattern is very broad**
The pattern `([%w_%./-]+%.[%w]+)` will match many non-path strings:
- URLs: `github.com/user/repo.git`, `api.openai.com`
- Method calls in prose: `Array.from`, `Object.keys`, `process.env`
- Dotted identifiers: `com.example.App`, `org.apache.Log4j`

**2. Backtick pattern too greedy**
Pattern `` `([^`]+%.[%w]+)` `` matches anything in backticks ending with a dot-extension:
- `config.timeout` - matches if file `config.timeout` exists
- `process.env` - `.env` looks like an extension
- `npm install package.json` - extracts trailing `package.json`

**3. Email addresses**
`user@example.com` - the `example.com` portion could match.

### Suggestion

Without the explicit `file://` prefix, there may be significant overhead from checking `file_exists()` on many false candidates, plus residual false positives when those strings happen to match existing files.

Consider whether the added complexity and edge cases are worth it vs. keeping the `file://` prefix requirement (perhaps just making it optional in the system prompt rather than removing it entirely). The file:// scheme provides unambiguous intent that a string is meant to be a file reference.

The tool part extraction (filePath from read/edit tools) is a nice addition regardless!

sudo-tee:
> ## False Positive Concerns
> Thanks for this refactor! Removing the system prompt dependency improves provider compatibility. However, I have concerns about false positives with the more permissive patterns:
> 
> ### Potential False Positives
> **1. Plain path pattern is very broad** The pattern `([%w_%./-]+%.[%w]+)` will match many non-path strings:
> 
> * URLs: `github.com/user/repo.git`, `api.openai.com`
> * Method calls in prose: `Array.from`, `Object.keys`, `process.env`
> * Dotted identifiers: `com.example.App`, `org.apache.Log4j`
> 
> **2. Backtick pattern too greedy** Pattern `` `([^`]+%.[%w]+)` `` matches anything in backticks ending with a dot-extension:
> 
> * `config.timeout` - matches if file `config.timeout` exists
> * `process.env` - `.env` looks like an extension
> * `npm install package.json` - extracts trailing `package.json`
> 
> **3. Email addresses** `user@example.com` - the `example.com` portion could match.
> 
> ### Suggestion
> Without the explicit `file://` prefix, there may be significant overhead from checking `file_exists()` on many false candidates, plus residual false positives when those strings happen to match existing files.
> 
> Consider whether the added complexity and edge cases are worth it vs. keeping the `file://` prefix requirement (perhaps just making it optional in the system prompt rather than removing it entirely). The file:// scheme provides unambiguous intent that a string is meant to be a file reference.
> 
> The tool part extraction (filePath from read/edit tools) is a nice addition regardless!

Thanks for the comments,

I understand your issue with false positives. and Will try to work on a better parsing solution. The file_exists is a fail safe on his own for most of the matches so I'm not to worry. 

The issue I have with only the file:// is staht it's to restrictive and requires a system prompt, and even with the system prompt it misses a lot of paths.



I will also ad as part of this MR a config for specifying a system prompt

sudo-tee:
@aweis89 

I did add more checks around urls, and better patterns to avoid false positives.

I also added a system prompt config.

I will merge this MR as it is right now and we can always update this later on if we see perf issues. 

I did some profiling on large outputs and it didn't impact the performance.

PR #201, opened 2026-01-24
refactor(keymap): Tweak default keymaps

### LINKED ISSUE #199: A little feedback about the default mappings
Feel free to close if you want, I'm happy to keep these in my configs, but a couple of the default mappings are a bit unintuitive.

Here's how I have mine configured:

```lua
{
  'sudo-tee/opencode.nvim',
  opts = {
    keymap = {
      input_window = {
        -- only submit with enter in normal mode, not insert mode
        ['<cr>'] = { 'submit_input_prompt', mode = { 'n' } },
        -- don't close the window on `<esc>`,
        -- too easily to close accidentally when trying to exit insert mode
        -- you can also just use :q
        ['<esc>'] = false,
      },
      output_window = {
        -- don't close the window on `<esc>`,
        -- too easily to close accidentally when trying to exit insert mode
        -- you can also just use :q
        ['<esc>'] = false,
      },
      session_picker = {
        -- don't override <C-n>, that should be reserved for "next entry",
        -- i.e. <C-n> and <C-p>
        new_session = false,
      },
    },
  },
},
```

mrjones2014:
cc @fredizzimo 

how does everyone feel about these minimal tweaks? 

Closes #199

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -64,7 +64,8 @@ M.defaults = {
       ['<leader>ods'] = { 'debug_session' },
     },
     input_window = {
-      ['<cr>'] = { 'submit_input_prompt', mode = { 'n', 'i' } },
+      ['<cr>'] = { 'submit_input_prompt', mode = { 'n' } },
+      ['<S-cr>'] = { 'submit_input_prompt', mode = { 'n', 'i' } },
       ['<esc>'] = { 'close' },
       ['<C-c>'] = { 'cancel' },
       ['~'] = { 'mention_file', mode = 'i' },
@@ -92,7 +93,7 @@ M.defaults = {
     session_picker = {
       rename_session = { '<C-r>' },
       delete_session = { '<C-d>' },
-      new_session = { '<C-n>' },
+      new_session = { '<C-s>' },
     },
     timeline_picker = {
       undo = { '<C-u>', mode = { 'i', 'n' } },


lua/opencode/config.lua
```diff
@@ -64,8 +63,8 @@ M.defaults = {
       ['<leader>ods'] = { 'debug_session' },
     },
     input_window = {
-      ['<cr>'] = { 'submit_input_prompt', mode = { 'n', 'i' } },
-      ['<esc>'] = { 'close' },
+      ['<cr>'] = { 'submit_input_prompt', mode = { 'n' } },
+      ['<S-cr>'] = { 'submit_input_prompt', mode = { 'i' } },
```
robfeldmann: For muscle-memory purposes, I would suggest enabling Shift-return in normal mode as well.

```suggestion
      ['<S-cr>'] = { 'submit_input_prompt', mode = { 'n', 'i' } },
```

mrjones2014: good idea :+1:

sudo-tee:
I am ok with changing the C-p and the enter for submitting (I changed it myself) 

But the escape is a pretty common pattern for closing panes  and I use it everyday.

mrjones2014:
I personally found myself closing it by accident way too easily with `<esc>`, but no worries, I'll keep that turned off in my own config. 

Updated PR as requested.

lua/opencode/config.lua
```diff
@@ -64,9 +63,10 @@ M.defaults = {
       ['<leader>ods'] = { 'debug_session' },
     },
     input_window = {
-      ['<cr>'] = { 'submit_input_prompt', mode = { 'n', 'i' } },
-      ['<esc>'] = { 'close' },
+      ['<cr>'] = { 'submit_input_prompt', mode = { 'n' } },
```
sudo-tee, line 67: I think you reverted the wrong line

mrjones2014: Oh, the code looks right but somehow the `<esc>` keymap got moved to a different order. Not sure how that happened but I will revert and fix it.

mrjones2014: Should be all good now.

PR #202, opened 2026-01-25
fix(core): only fail startup on actual errors

guill:
The latest version of OpenCode (1.1.35) outputs a log line to stderr on startup (see https://github.com/anomalyco/opencode/issues/10506 ). This causes the NeoVim plugin to hang indefinitely.

This commit makes two changes:
1. If we fail to start the OpenCode server, we display a notification with the error rather than hanging indefinitely.
2. We don't allow DEBUG, INFO, or WARN-level messages to fail initialization.

### DIFF
--- a/lua/opencode/core.lua
+++ b/lua/opencode/core.lua
@@ -85,7 +85,17 @@ M.open = Promise.async(function(opts)
     ui.focus_output({ restore_position = are_windows_closed })
   end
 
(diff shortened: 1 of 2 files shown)


sudo-tee:
Thanks for the PR 

I am away for the weekend but it looks good

I will merge so it unblock users with the latest opencode version

PR #206, opened 2026-01-26
refactor(question-window): replace question picker

### LINKED ISSUE #197: "ask" tool has sub optimal UX
When the question is longer than just a very short question, the answer UX is suboptimal. I'm using the default Snacks picker.

Screenshot attached below. See how the question appears in a notification (I'm using `nvim-notify`); notifications dismiss on a timer, and you can't then go reference the question again without closing the picker, which cancels the question and terminates the prompt.

Image

I'm not totally sure what the best solution is here. Maybe custom UI? Or maybe we could use [nui.nvim](https://github.com/MunifTanjim/nui.nvim) components?

I'm willing to work on this myself and submit a PR, but I'd like some input from you on the direction you'd like to go for your plugin.

sudo-tee:
- Replace question picker with a custom dialog in the chat window

Question window
image

Permission window
image

This should close  #197


mrjones2014:
This works great thank you!

PR #207, opened 2026-01-27
fix(context): simplify file addition logic

aweis89:
- These checks are already present [upstream](https://github.com/sudo-tee/opencode.nvim/blob/main/lua/opencode/context.lua#L88) in the call stack.

- Primary motivation is that I wanted to be able to add files outside the current working directory to the context. This enables clients to do so by using `require("opencode.context").ChatContext.add_file`bypassing these checks instead of `require("opencode.context").add_file` which still won't allow it.


sudo-tee:
Thanks for the PR

I am fine with this

Issue #205, opened 2026-01-26
`/` won't trigger commands after a command has already been executed


PowerUser64:
Quite often I like to execute commands like switch model (`/model`) and then switch agent (`/agent`). For some reason when I try to switch model and then switch agent, the second slash command never works on the first try and I have to open and close the I to get it to work. This seems like it's related to #130.

**Steps to reproduce:** (described above a bit)
1. Run `/model`, switch models
2. Run `/agent` -- the completion menu for agents won't pop up and you'll instead get the file path completion menu

sudo-tee:
@PowerUser64 

I am not able to reproduce the issue, what completion system are you using blink or cmp ?

Can you provide a screenshot or screen recording of the issue ?

Are you sure there is no character before the / in the input window such as a space ?

bowentan:
Came across the same issue: after selecting a command popped by `blink`, delete it and again type `/` will not pop up the completion.

https://github.com/user-attachments/assets/f83367e0-8c9b-40b7-a04f-24d3e87e4bed

Related env:

- macOS 13.7.8
- wezterm
- nvim 0.11

sudo-tee:
@bowentan what shortcut are you using to insert /agent_init ?  this command execute immediately and not insert itself in the input window.

bowentan:
@sudo-tee Used the default confirm key `<C-y>`.

PowerUser64:
Yeah, this is basically what I have going on. I am using blink.cmp just like @bowentan, although I have `<C-l>` (lowercase L) set as my confirm key. I could record the same video as is posted above, only when I type slash the second time, the path completion source will appear.

sudo-tee:
@bowentan @PowerUser64 

I created a MR to try to address the issue. I am not able to reproduce it reliably.

https://github.com/sudo-tee/opencode.nvim/pull/213

If you have a couple of minutes to try my fix it would be greatly appreciated.

Thanks

inferst:
@sudo-tee I had the same issue

checked your commit and it looks like it is fixed

PowerUser64:
Works for me too! Thanks @sudo-tee :)

PR #213, opened 2026-01-29
fix(blink): improve provider override logic and path completion handling

sudo-tee:
This should improve #205


inferst:
Faced with some wrong behavior in the commit.

For example, if I press the spacebar in the middle of a command and then press backspace and continue typing the command, then autocompletion doesn't work.

sudo-tee:
> Faced with some wrong behavior in the commit.
> 
> For example, if I press the spacebar in the middle of a command and then press backspace and continue typing the command, then autocompletion doesn't work.

I will have another look, 
This is a complex issue as the completion is not opened anymore in this case

PR #215, opened 2026-01-29
feat(ui): only auto hide input if contents empty

disrupted:
I really like the auto-hide feature for the input but I wish it would stay open while drafting a new prompt. It's easy to forget what you were composing if the window is hidden. that's why I added a condition to check if it's empty before hiding the window. let me know what you think


sudo-tee:
It makes sense to me

Thanks for the PR

PR #217, opened 2026-01-29
feat(ui): auto-resize input win height

disrupted:
this is just a draft implementation. seeing as the Opencode TUI auto-resizes its prompt I wanted to see if we could add a similar feature to this plugin. curious to hear your thoughts on this.

demo

```lua
{
  ui = {
    input = {
      min_height = 0.10,
      max_height = 0.25,
      text = {
        wrap = true,
      },
      auto_hide = true,
    }
  }
}
```

https://github.com/user-attachments/assets/2a68207e-14b8-4fb6-ae0f-7e5644738a32

### DIFF
--- a/lua/opencode/ui/dialog.lua
+++ b/lua/opencode/ui/dialog.lua
@@ -18,7 +18,7 @@
 
 ---@class Dialog
 ---@field private _config DialogConfig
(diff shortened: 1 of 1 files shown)


sudo-tee:
Thanks I really like this :)

PR #218, opened 2026-01-30
fix(dialog): clean up keymaps when dialog closes

guill:
The Dialog class was failing to clean up its keymaps (`j`, `k`, `<Up>`, `<Down>`, `<CR>`, `<Esc>`, `1-9`) when torn down. This caused keymaps to accumulate on the output buffer, leaving it in an inconsistent state that could interfere with other keymaps including `<C-c>` for cancel and `j`/`k` for moving up and down.

Root cause: vim.keymap.set() returns true/nil, not a keymap identifier. The code was storing these useless return values and then trying to use them as key names in vim.keymap.del(), which silently failed.

Fix: Store the actual key binding strings (e.g., `"j"`, `"<CR>"`) instead of the return value, then use those strings to properly delete the keymaps during cleanup. Also added a buffer validity check to prevent errors when cleaning up after the buffer has been deleted.

### DIFF
--- a/lua/opencode/ui/dialog.lua
+++ b/lua/opencode/ui/dialog.lua
@@ -18,7 +18,7 @@
 
 ---@class Dialog
 ---@field private _config DialogConfig
----@field private _keymaps integer[] List of keymap IDs for cleanup
+---@field private _keymaps string[] List of key bindings for cleanup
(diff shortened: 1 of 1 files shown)


sudo-tee:
Thanks for the fix.

PR #219, opened 2026-01-31
feat(event_manager): fire User autocmds for server-sent events

disrupted:
coming from the other opencode.nvim plugin, this was something that I missed. currently we only have a select few hooks that the user can leverage.

with this PR we emit all server-side events from Opencode as a custom User autocmd. I find it generally useful when plugins provide these to trigger custom logic. It can also be used to integrate with other plugins and enable advanced workflows.

simple example from my config: I set up custom keymaps only when Opencode requests permissions, then clear those up after replying.

### DIFF
--- a/lua/opencode/event_manager.lua
+++ b/lua/opencode/event_manager.lua
@@ -100,10 +100,9 @@ local util = require('opencode.util')
 --- @field properties {ide: string}
 
 --- @class ServerStartingEvent
---- @field server_job table
+--- @field url string
 
 --- @class ServerReadyEvent
---- @field server_job table
 --- @field url string
 
 --- @class ServerStoppedEvent
@@ -323,14 +322,12 @@ function EventManager:emit(event_name, data)
     end
   end
 
-  if not vim.startswith(event_name, 'custom.') then
-    vim.api.nvim_exec_autocmds('User', {
-      pattern = 'OpencodeEvent:' .. event_name,
-      data = {
-        event = event,
-      },
-    })
-  end
+  vim.api.nvim_exec_autocmds('User', {
+    pattern = 'OpencodeEvent:' .. event_name,
+    data = {
+      event = event,
+    },
+  })
 end
 
 --- Start the event manager and begin listening to server events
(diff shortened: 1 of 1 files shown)


lua/opencode/event_manager.lua
```diff
@@ -303,26 +303,36 @@ end
...
-    if not ok then
-      vim.notify('Error calling ' .. event_name .. ' listener: ' .. result, vim.log.levels.ERROR)
+      if not ok then
+        vim.notify('Error calling ' .. event_name .. ' listener: ' .. result, vim.log.levels.ERROR)
+      end
     end
   end
+
+  -- Fire User autocmd for external subscribers
+  if not vim.startswith(event_name, 'custom.') then
```
sudo-tee: Why filter out custom events, some are useful  like `custom.server_ready` in the casse you want to grab the server url ?

lua/opencode/event_manager.lua
```diff
@@ -303,26 +303,36 @@ end
...
 
   local event = { type = event_name, properties = data }
 
   if config.debug.capture_streamed_events then
     table.insert(self.captured_events, vim.deepcopy(event))
   end
 
-  for _, callback in ipairs(listeners) do
-    local ok, result = util.pcall_trace(callback, data)
+  -- Dispatch to internal listeners
```
sudo-tee: No need for comments like this when the code is obvious

tests/unit/event_manager_spec.lua
```diff
@@ -114,8 +114,69 @@ describe('EventManager', function()
...
   end)
-end)
\ No newline at end of file
+
+  describe('User autocmd events', function()
+    it('should fire User autocmd when emitting events', function()
+      local autocmd_called = false
+      local autocmd_data = nil
+
+      -- Set up autocmd listener
```
sudo-tee: Same for theses comments..

lua/opencode/event_manager.lua
disrupted: I agree, but I believe it would require sanitizing the event data before we pass it to `exec_autocmds`. right now Neovim throws an error for some of the `custom.<>` events because the table contains arbitrary types (userdata, Lua functions) which apparently cannot be serialized. https://github.com/sudo-tee/opencode.nvim/blob/4b950e091634b9f9469a7fa28066957c74645969/lua/opencode/event_manager.lua#L343
I am not familiar with best-practices how to deal with that. do you have a proposal?

disrupted: I tried removing the userdata from the custom events (only passing the server url now). not sure if this will break anything. let me know in case you have a preference how to handle it.

sudo-tee: It should be fine, the actual event data is not used

PR #220, opened 2026-02-01
fix(ui): fix tip of cancel keymap in footer

v1nh1shungry:
Problem: Wrong (maybe legacy i guess) name when getting the cancel keymap. This causes the tip in the footer is always `<C-c>` even if I set another key.


sudo-tee:
Thanks for this fix.

PR #222, opened 2026-02-01
fix(context): deduplicate selection ranges

disrupted:
prevent duplicate selection from being added to context. allow overlapping line ranges, block if exact same line range already exists in context


sudo-tee:
@disrupted 

You are on fire.

Thanks for another fix

PR #223, opened 2026-02-01
fix(context): allow selection without `current_file` option enabled

disrupted:
adding selection ranges to the context (e.g. using the `focus_input` keymap in visual mode) currently depends on `context.current_file` being enabled.
unless I misunderstood something, I believe this coupling is unintentional and might have been added by mistake. passing the entire current file as context defeats the purpose of the targeted selection (results in unnecessary context bloat)

no changes to quick chat as it already works on selection without passing the whole file/buffer as context

### DIFF
--- a/lua/opencode/context/base_context.lua
+++ b/lua/opencode/context/base_context.lua
@@ -149,6 +149,20 @@ function M.get_current_file(buf, context_config)
 end
 
 ---@param buf integer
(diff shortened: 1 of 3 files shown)


sudo-tee:
You are right the coupling is not intentional, the selection use the 'current' file only to get the filename and the language

Thanks for the fix

PR #224, opened 2026-02-01
fix(ui): preserve zoom state on redraw

disrupted:
When using the zoomed agent chat window (via `:Opencode toggle_zoom`), the window width would reset to normal after certain actions (like responding to permission prompts), but the internal zoom state (`state.pre_zoom_width`) remained set. This required calling `toggle_zoom` twice to get back to zoomed width. Now the zoom width is preserved across all UI actions unless the user actively toggles back to normal width.


sudo-tee:
@disrupted 
Thanks for the fix 

This looks good to me 

Once the merge conflict is fixed I will merge it

PR #227, opened 2026-02-02
fix(ui): show input when adding context

disrupted:
With `auto_hide=true` when adding new context (shell command output, file picker, etc.) the input window didn't show. this can be confusing and users might wonder where their content went. it definitely happened to me while I was trying out the shell command feature for the first time. I was staring at a blank session view and for a hot minute I thought it was broken because nothing seemed to happen 😄


sudo-tee:
This makes sense to me

Thanks for the PR

PR #229, opened 2026-02-03
Currently splitting implementation via new PRs - feat(ui): preserve buffers on toggle

jensenojs:
## Background

This PR improves UI toggle behavior by introducing `ui.persist_state` (default: `true`) and refining window and buffer lifecycle handling.

Goal: when toggling UI off and on, preserve user context (buffers and cursor or scroll intent) without breaking close semantics.

## Behavior Changes

- Add `ui.persist_state` (default `true`):
  - `toggle` now hides windows and preserves buffers when enabled.
  - when disabled, behavior remains full close (buffers are deleted).

- Normalize window state semantics:
  - `visible`: input or output window is currently visible
  - `hidden`: windows are closed but restorable hidden buffers exist
  - `closed`: no visible windows and no restorable hidden buffers

- Fix toggle open-state detection:
  - output-only view (input auto-hidden) is treated as `visible` (not `closed`).

- Make `get_window_state()` read-only:
  - remove cursor-saving side effects from getter.

- Restore policy after hidden reopen:
  - if user was following bottom before hide, allow sync-to-latest behavior on reopen;
  - if user was reading in the middle, preserve reading position.

- Explicit hidden-buffer lifecycle:
  - stash / consume / clear hidden snapshot
  - validity checks for restored buffers

- Type and LSP cleanup:
  - reduce false diagnostics while keeping `gd` navigation targeting real implementations.

## Test Coverage

Added and updated behavior-focused tests:

- `tests/unit/persist_state_spec.lua`
  - config default and override
  - preserve on hide plus reopen reuse
  - closed, visible, hidden state transitions
  - `get_window_state()` purity (no mutation)
  - output-only visible semantics
  - `persist_state=false` full-close path

- `tests/unit/core_spec.lua`
  - hidden restore render gating:
    - no forced full render when preserving reading position
    - full render allowed when previously at bottom

## Reviewer Guide

Suggested review order:

1. `lua/opencode/state.lua`: state model plus hidden snapshot lifecycle
2. `lua/opencode/ui/ui.lua`: close and create window lifecycle split
3. `lua/opencode/core.lua`: reopen render gating behavior
4. `lua/opencode/api.lua`: toggle semantics plus getter purity
5. tests for behavior lock-in

## Risks / Compatibility

- Default behavior changes toward preservation (`persist_state=true`).
- Compatibility path is preserved via `persist_state=false`.
- Main risk area is state transitions (`visible/hidden/closed`), covered by regression tests above.

### DIFF
--- a/lua/opencode/api.lua
+++ b/lua/opencode/api.lua
@@ -18,23 +18,37 @@ local M = {}
 ---Get the current window state of opencode
 ---@return {status: 'closed'|'hidden'|'visible', visible: boolean, hidden: boolean, position: string, windows: OpencodeWindowState|nil, cursor_positions: {input: integer[]|nil, output: integer[]|nil}}
 function M.get_window_state()
+  ---@type OpencodeWindowState|nil
   local windows = state.windows
   local status = state.get_window_status()
 
-  local cursor_positions = {
-    input = nil,
-    output = nil,
-  }
-
-  -- If windows are visible, get current cursor positions
-  if status == 'visible' then
-    cursor_positions.input = state.save_cursor_position('input', windows.input_win)
-    cursor_positions.output = state.save_cursor_position('output', windows.output_win)
-  else
-    -- Return saved positions
-    cursor_positions.input = state.get_cursor_position('input')
-    cursor_positions.output = state.get_cursor_position('output')
+  ---@param win_id integer|nil
+  ---@return integer[]|nil
+  local function get_window_cursor(win_id)
+    if not win_id or not vim.api.nvim_win_is_valid(win_id) then
+      return nil
+    end
+    local ok, pos = pcall(vim.api.nvim_win_get_cursor, win_id)
+    if ok then
+      return pos
+    end
+    return nil
   end
+
+  local input_cursor = get_window_cursor(windows and windows.input_win)
+  if input_cursor == nil then
+    input_cursor = state.get_cursor_position('input')
+  end
+
+  local output_cursor = get_window_cursor(windows and windows.output_win)
+  if output_cursor == nil then
+    output_cursor = state.get_cursor_position('output')
+  end
+
+  local cursor_positions = {
+    input = input_cursor,
+    output = output_cursor,
+  }
 
   return {
     status = status,
@@ -94,43 +108,50 @@ end
 --- Check if opencode windows are in the current tab page
 --- @return boolean
 local function are_windows_in_current_tab()
-  if not state.windows or not state.windows.output_win then
+  if not state.windows then
     return false
   end
 
   local current_tab = vim.api.nvim_get_current_tabpage()
-  local ok, win_tab = pcall(vim.api.nvim_win_get_tabpage, state.windows.output_win)
-  return ok and win_tab == current_tab
+
+  local function is_window_in_current_tab(win_id)
+    if not win_id or not vim.api.nvim_win_is_valid(win_id) then
+      return false
+    end
+    local ok, win_tab = pcall(vim.api.nvim_win_get_tabpage, win_id)
+    return ok and win_tab == current_tab
+  end
+
+  return is_window_in_current_tab(state.windows.input_win)
+    or is_window_in_current_tab(state.windows.output_win)
 end
 
 M.toggle = Promise.async(function(new_session)
-  -- When auto_hide input is enabled, always focus input; otherwise use last focused
   local focus = 'input' ---@cast focus 'input' | 'output'
   if not config.ui.input.auto_hide then
     focus = state.last_focused_opencode_window or 'input'
   end
 
-  -- Check if windows are actually open (not just if state.windows exists)
-  -- When hidden with persist_state, state.windows exists but window IDs are nil
+  local function is_valid_win(win_id)
+    return win_id ~= nil and vim.api.nvim_win_is_valid(win_id)
+  end
+
   local windows_open = state.windows ~= nil
-    and state.windows.input_win ~= nil
-    and vim.api.nvim_win_is_valid(state.windows.input_win)
+    and (is_valid_win(state.windows.input_win) or is_valid_win(state.windows.output_win))
     and are_windows_in_current_tab()
 
   if not windows_open then
-    -- Windows closed (or hidden with preserved buffers), open/restore them
-    -- Note: When hidden with persist_state, state.windows exists but window IDs are nil.
-    -- We should not call M.close() here as it would delete the preserved buffers.
-    -- Just clear the stale window state reference; create_windows will handle hidden buffers.
     if state.windows and not state.has_hidden_buffers() then
-      -- No hidden buffers, safe to clear the stale state
       state.windows = nil
     end
     core.open({ new_session = new_session == true, focus = focus, start_insert = false }):await()
   else
-    -- Windows open, close/hide them
+    if state.display_route then
+      M.close()
+      return
+    end
+
     if config.ui.persist_state then
-      -- Keep buffers and avoid renderer teardown for faster restore.
       M.hide()
     else
       M.close()
--- a/lua/opencode/ui/ui.lua
+++ b/lua/opencode/ui/ui.lua
@@ -1,4 +1,6 @@
+---@type OpencodeConfig & OpencodeConfigModule
 local config = require('opencode.config')
+---@type OpencodeStateModule
 local state = require('opencode.state')
 local renderer = require('opencode.ui.renderer')
 local output_window = require('opencode.ui.output_window')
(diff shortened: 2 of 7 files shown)


disrupted:
great addition! this is so useful, imo it would make for a sensible default. there is still `:Opencode close` for when the user would want to actually close it

lua/opencode/config.lua
```diff
@@ -109,6 +109,7 @@ M.defaults = {
     position = 'right',
     input_position = 'bottom',
     window_width = 0.40,
+    keep_buffers_on_toggle = false,
```
sudo-tee: don't forget to add this to the README.md

lua/opencode/api.lua
```diff
@@ -51,6 +51,11 @@ function M.close()
   ui.close_windows(state.windows)
 end
 
+-- Hide the UI but keep buffers for fast restore.
+function M.hide()
```
sudo-tee, line 100: I would add a command also `Opencode hide` what do you think ?

sudo-tee:
This sems like a useful feature.

As I will have to maintain this in the future, I would like it if you add a couple of tests for this.

As well as updating the README file

Thanks a lot for the PR

jensenojs:
Sorry, I found some more subtle issues. 

I'm glad you reviewed so carefully, including that it should probably be the default behavior. I will resolve all the above issues as soon as possible.

lua/opencode/api.lua
jensenojs: done

jensenojs:
I am quite satisfied with my local behavior, (my nvim configuration will switch when the position is both left and center, plus I want to retain the cursor position when toggling, so although the code is relatively complex, I also spent quite a while calibrating through actual testing. I might record a video hhh)

The overall code writing and cleanup were all done by kimi2.5. There may be some unnecessary comments or still overdesigned parts here. I will carefully review my PR tomorrow and clean up that junk (it's late night here now, I have to sleep). I pushed the code up first in case someone might want to try it out early. @sudo-tee @disrupted 

---

update : 

https://github.com/user-attachments/assets/910f98c6-02e1-493f-b4fd-bd6d837ea951

jensenojs:
Quick reviewer note: this PR is primarily a behavior boundary fix, not a refactor.

Please focus on:
- state semantics (`visible/hidden/closed`)
- hide and reopen lifecycle (`stash/consume/clear`)
- reopen policy split (follow-bottom vs preserve-reading-position)

Most important files:
- `lua/opencode/state.lua`
- `lua/opencode/ui/ui.lua`
- `lua/opencode/core.lua`
- `lua/opencode/api.lua`
- `tests/unit/persist_state_spec.lua`

disrupted:
at 0:13 in the video it looks like the cursor jumped up a little when re-opening the output buffer. do you know why that's happening?

sudo-tee:
Thanks for the PR, 

I will play with it a little bit in the coming days as welle a redo a finale review.

jensenojs:
> at 0:13 in the video it looks like the cursor jumped up a little when re-opening the output buffer. do you know why that's happening?在视频的 0:13 处，重新打开 output buffer 时光标似乎稍微向上跳了一下。你知道这是为什么吗？

When the cursor bounces back from the editing window to the opencode output window, if it wasn't at the bottom before (for example, if I was looking at messages in the middle), I want it to reappear at the corresponding position.

Actually, it's not just the cursor jumping up; it seems to be one of my autojump commands that centers the entire window, although it does create a bit of visual noise.

jensenojs:
Need a more detailed performance evaluation and bottleneck analysis (in progress)

sudo-tee:
I played a little bit with it this morning, 

It progressing great, I like that I  don't have to wait for a full render if I toggle the panel.

I find one issue where if you move the cursor in the outpout window withoud scrolling and close/reopen, the cursor will be set at the bottom instead of restoring to the last position.

There is also some tests failiong in the branch.

jensenojs:
> I played a little bit with it this morning,
> 
> It progressing great, I like that I don't have to wait for a full render if I toggle the panel.
> 
> I find one issue where if you move the cursor in the outpout window withoud scrolling and close/reopen, the cursor will be set at the bottom instead of restoring to the last position.
> 
> There is also some tests failiong in the branch.

I'm sorry—I've actually been running local tests in my off hours these past few days. What you mentioned is a known issue to me, but at the same time there's a more serious problem: using end-to-end load testing I can reproduce a severe performance issue. Restoring a cached buffer requires reattaching the window and setting the filetype, which triggers Treesitter to reparse all existing content in full. That cost negates the benefits of avoiding re-rendering.

I can't come up with a complete, well-reasoned conclusion right now, and I don't want to treat this project with a restless attitude. I'll need to look through the code myself again tomorrow — I can't keep relying on GPT-5.3 and Kimi2.5.

At first I implemented this feature almost entirely by vibe coding; when reviewing the code I looked at the data structures and added a mock to create a performance comparison test, which deceptively showed a 40% performance improvement.

But when I later compared latency end-to-end in a real opencode session (looped 100 times, comparing mean and std), the results were terrible (I really should've picked up on this script from the start). I was able to identify some plausible issues, like duplicate subscriptions (possibly a legacy problem) or treesitter doing full parses that took too long... The final optimizations weren't particularly satisfying, and since I didn't have deeper control over the code during that time, it was pretty bad.

jensenojs:
Performance testing details are as follows, along with my originally intended commit message:

图片

---

    Added the `persist_state` configuration option to allow buffer states to be maintained on close/switch for quick recovery. Includes the following improvements:
    
    - Added the `ui.persist_state` configuration (default true) to control closing behavior
    - Implemented a hide/restore mechanism that preserves cursor positions, scroll states, and window layouts
    - Add new APIs such as `api.hide()` and `api.get_window_state()`
    - Optimize the event manager by adding health checks and external message polling
    - Add performance benchmarking scripts to demonstrate performance improvements from persistence
    - Update documentation to include usage examples and three-state layout conversion solutions

---

However, this commit has too many changes. For maintainability, I'd like to split it into several smaller PRs. Is that possible? (I'll assume you've agreed (lol @sudo-tee 

I found I don't like that visual noise either, so I also added an option to control whether this thing can be turned on. @disrupted

sudo-tee:
If you are able to split in multiple PR . I will agree. It will be easier

lua/opencode/ui/input_window.lua
```diff
@@ -480,6 +484,9 @@ function M.setup_autocmds(windows, group)
     group = group,
     buffer = windows.input_buf,
     callback = function()
+      -- Save cursor position before leaving
```
sudo-tee, line 487: If possible, I'd like to keep comments only for explaining complex snippet.

AI tends to add comments for abvious thinks like this.

lua/opencode/ui/ui.lua
```diff
@@ -76,22 +111,28 @@ local function open_split(direction, type)
 end
 
 function M.create_split_windows(input_buf, output_buf)
-  if state.windows then
-    M.close_windows(state.windows)
+  if state.windows and state.windows.input_win and vim.api.nvim_win_is_valid(state.windows.input_win) then
```
sudo-tee, line 114: There is a function in each window module calles M.mounted().

You might want to leverage it instead, as it checks the same conditions as this

lua/opencode/ui/input_window.lua
jensenojs: will do

lua/opencode/api.lua
```diff
@@ -15,6 +15,51 @@ local Promise = require('opencode.promise')
...
+      return nil
+    end
+    local ok, pos = pcall(vim.api.nvim_win_get_cursor, win_id)
+    if ok then
+      return pos
+    end
+    return nil
+  end
+
+  local input_cursor = get_window_cursor(windows and windows.input_win)
```
sudo-tee, line 38: you could simplify the nil checks with this

```lua
local input_cursor = get_window_cursor(windows and windows.input_win) or  state.get_cursor_position('input')

local output_cursor = get_window_cursor(windows and windows.output_win) or state.get_cursor_position('output')
```

lua/opencode/api.lua
```diff
@@ -51,36 +96,66 @@ function M.close()
...
 
-  if state.windows == nil or not are_windows_in_current_tab() then
-    if state.windows then
-      M.close()
+  local function is_valid_win(win_id)
+    return win_id ~= nil and vim.api.nvim_win_is_valid(win_id)
+  end
+
+  local windows_open = state.windows ~= nil
+    and (is_valid_win(state.windows.input_win) or is_valid_win(state.windows.output_win))
```
sudo-tee, line 140: You could use both `input_window.mounted()`  and `input_window.mounted()` here

PR #230, opened 2026-02-03
perf: optimize string concatenation and namespace caching

jinzhongjia:
Summary
- Replace string concatenation in loops with `table.concat` in ID generation
- Cache namespace ID at module level instead of repeated creation
 Changes
 `lua/opencode/id.lua`
- `random_base62()`: Use table + `table.concat` instead of `result = result .. char` (O(n) vs O(n²))
- `generate_new_id()`: Same optimization for timestamp byte extraction
 `lua/opencode/ui/input_window.lua`
- Cache `placeholder_ns` at module level
- Remove 3 redundant `nvim_create_namespace('input_placeholder')` calls
 Why
String concatenation in Lua creates new string objects each iteration, causing O(n²) complexity. While the impact is minimal for short strings (26 chars), this follows better Lua practices.
Namespace IDs are stable for the same name, but caching avoids unnecessary function calls on every placeholder refresh.

### DIFF
--- a/lua/opencode/id.lua
+++ b/lua/opencode/id.lua
@@ -18,14 +18,14 @@ local LENGTH = 26
 -- Generate random base62 string
 local function random_base62(length)
   local chars = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
-  local result = ''
+  local parts = {}
 
-  for _ = 1, length do
+  for i = 1, length do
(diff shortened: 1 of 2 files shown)


jinzhongjia:
This is a little optimize, it will increase our performance a bit

sudo-tee:
The id calss is rarely used. I think it's only used for the init command.

But I am ok with it being more optimized. It was a loose port from the opencode implementation in typescript

Thanks for the PR

PR #232, opened 2026-02-03
fix(ui): escape filepath to prevent wrong formatting

disrupted:
context files with special filenames could falsly be interpreted as markdown formatting syntax

before: **init**.py

now: `__init__.py`


sudo-tee:
This looks good to me.

Thanks for taking the time to regenerate the tests snapshote :)

PR #233, opened 2026-02-03
fix(ui): prevent duplicate windows

### LINKED ISSUE #144: Got two input buffers when I resize the nvim window
When I resize the neovim window, I got two chatting input buffers .

Image

disrupted:
fix #144
not entirely sure if this approach aligns with best-practices
at least I am no longer seeing the issue on my end

### DIFF
--- a/lua/opencode/ui/autocmds.lua
+++ b/lua/opencode/ui/autocmds.lua
@@ -75,15 +75,18 @@ function M.setup_autocmds(windows)
   end
 end
 
+---@param windows OpencodeWindowState?
 function M.setup_resize_handler(windows)
   local resize_group = vim.api.nvim_create_augroup('OpencodeResize', { clear = true })
   vim.api.nvim_create_autocmd('VimResized', {
     group = resize_group,
     callback = function()
+      require('opencode.ui.ui').reconcile_windows(windows, 'input')
+      require('opencode.ui.ui').reconcile_windows(windows, 'output')
       require('opencode.ui.topbar').render()
       require('opencode.ui.footer').update_window(windows)
-      require('opencode.ui.input_window').update_dimensions(windows)
-      require('opencode.ui.output_window').update_dimensions(windows)
+      input_window.update_dimensions(windows)
+      output_window.update_dimensions(windows)
     end,
   })
 
--- a/lua/opencode/ui/output_window.lua
+++ b/lua/opencode/ui/output_window.lua
@@ -23,6 +23,7 @@ function M._build_output_win_config()
   }
 end
 
+---@param windows OpencodeWindowState?
 function M.mounted(windows)
   windows = windows or state.windows
   return windows and windows.output_buf and windows.output_win and vim.api.nvim_win_is_valid(windows.output_win)
--- a/lua/opencode/ui/ui.lua
+++ b/lua/opencode/ui/ui.lua
@@ -296,4 +296,48 @@ function M.toggle_zoom()
   end
 end
 
+---Reconcile window state when duplicate windows may have been created
+---This can happen when external commands like 'tabdo' manipulate window layouts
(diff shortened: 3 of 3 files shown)


sudo-tee:
I played a little bit wit it.
It seems to work when resizingf but it does not fix the root cause.

I am still able to duplicate the input window if I wun `:tabdo wincmd =` when focused in the input.

I will play a little bit with this to see if there is a better solution

sudo-tee:
How about a more generic approach with a utility for fix a buf to a window, it's like the opposite of winfixbuf.

I came up with the following based on your reconcile_windows

```lua
--- Prevents a buffer from appearing in multiple windows (opposite of 'winfixbuf')
---
--- This module solves the problem where buffers can appear in multiple windows
--- despite having 'winfixbuf' set. The 'winfixbuf' option prevents a window from
--- changing buffers, but doesn't prevent a buffer from appearing in multiple windows.
---
local M = {}
local buff_to_win_map = {}
local global_autocmd_setup = false

local function close_duplicates(buf, get_win)
  local intended = get_win()
  if not intended then
    return
  end

  local wins = vim.fn.win_findbuf(buf)
  if #wins <= 1 then
    return
  end

  for _, win in ipairs(wins) do
    if win ~= intended and vim.api.nvim_win_is_valid(win) then
      vim.schedule(function()
        pcall(vim.api.nvim_win_close, win, true)
      end)
    end
  end
end

local function setup()
  if global_autocmd_setup then
    return
  end
  global_autocmd_setup = true

  vim.api.nvim_create_autocmd('WinNew', {
    callback = function()
      vim.schedule(function()
        for buf, get_win in pairs(buff_to_win_map) do
          if vim.api.nvim_buf_is_valid(buf) then
            close_duplicates(buf, get_win)
          else
            buff_to_win_map[buf] = nil
          end
        end
      end)
    end,
  })
end

--- Protect a buffer from appearing in multiple windows
---@param buf integer Buffer number
---@param get_intended_window fun(): integer? Function returning intended window ID
function M.fix_to_win(buf, get_intended_window)
  setup()

  buff_to_win_map[buf] = get_intended_window
  vim.api.nvim_create_autocmd('BufWinEnter', {
    buffer = buf,
    callback = function()
      close_duplicates(buf, get_intended_window)
    end,
  })
end

return M
```

And it could be used like so on a buffer after creation

```lua
  local buffixwin = require('opencode.ui.buf_fix_win')
  buffixwin.fix_to_win(input_buf, function()
    local state = require('opencode.state')
    return state.windows and state.windows.input_win
  end)
```

In my quick tests it covers, 

resizing, `tabdo wincmd=`, `vsplit and split` from messing the layout.

disrupted:
great, clearly this is a much more robust solution! feel free to open a PR to supersede mine. 

overall I haven't fully understand why this issue occurs in the first place. it feels like we're doing something fundamentally wrong that makes these workarounds necessary. but I am happy about any pragmatic solution at this point. we can always refine it later once we identify the root cause of this problem

sudo-tee:
> great, clearly this is a much more robust solution! feel free to open a PR to supersede mine.
> 
> overall I haven't fully understand why this issue occurs in the first place. it feels like we're doing something fundamentally wrong that makes these workarounds necessary. but I am happy about any pragmatic solution at this point. we can always refine it later once we identify the root cause of this problem

I have tested other similar plugins and most of them have similar issues .  you can split , duplicate  windows. I think it's more of an issue related to fixed window layout,  some plugins like  https://github.com/stevearc/stickybuf.nvim uses a similar approach.

I will create a PR to use my fix

Thanks

sudo-tee:
Closing this one in favor of 
https://github.com/sudo-tee/opencode.nvim/pull/238

Which was inspired by the work on this branch

PR #234, opened 2026-02-03
fix(formatter): format optional grep args

disrupted:
this prevents optional grep arguments from showing up as `` (empty string in backticks). render-markdown does not like them either


sudo-tee:
Thanks for the fix.

PR #235, opened 2026-02-04
fix(blink): use builtin kinds when possible

AbaoFromCUG:
* `Folder` and `File` are builtin kind which is supported by `blink.cmp`, only need `kind` property for this case
* Custom kind need pass `kind_name` and `kind_icon`, e.g. `Subagent` and  `󰚩 `


sudo-tee:
Thanks for the PR

I am fine wit this change

PR #243, opened 2026-02-05
fix(renderer): handle session title updates

### LINKED ISSUE #242: new session does not get updated with automatic title
when starting a new unnamed session, opencode generates an automatic title for it based on the context of the first user message. the server sends it as a `session.updated` event. it is however not updated in this plugin. currently it requires restarting nvim (or the plugin) to see the new title.

looking at the captured events more closely the new title is actually part of 2 events

```jsonl
{
  "type": "message.updated",
  "properties": {
    "info": {
      "model": { "providerID": "openai", "modelID": "gpt-5.3-codex" },
      "sessionID": "ses_3d0b471aeffeLvW9F7ux3Ol5im",
      "id": "msg_c2f4c6c00001wyAVBQKhEeNV36",
      "summary": { "title": "Branch review request", "diffs": [] },
      "role": "user",
      "agent": "build",
      "time": { "created": 1770320063488 }
    }
  }
},
{ "type": "custom.emit_events.finished", "properties": [] },
{ "type": "custom.emit_events.started", "properties": [] },
{
  "type": "session.updated",
  "properties": {
    "info": {
      "version": "1.1.52",
      "directory": "/Users/disrupted/.local/share/nvim/lazy/opencode-native",
      "slug": "proud-star",
      "id": "ses_3d0b471aeffeLvW9F7ux3Ol5im",
      "title": "Branch review request",
      "projectID": "29d43526f88157cd4edd071b899dd01f240b771b",
      "time": { "created": 1770320006737, "updated": 1770320063913 },
      "summary": { "additions": 0, "deletions": 0, "files": 0 }
    }
  }
}
```

I am gonna have a look and see whether I can create a PR for it.

disrupted:
fix #242

each `session.updated` event seems to be a patch-like delta update. now we merge fields into existing session state and update the session title if it changed.

### DIFF
--- a/lua/opencode/ui/renderer.lua
+++ b/lua/opencode/ui/renderer.lua
@@ -851,9 +851,9 @@ function M.on_session_updated(properties)
 
   local merged_session = vim.tbl_deep_extend('force', vim.deepcopy(current_session), updated_session)
 
-  -- mutate existing `state.active_session` table in place
-  -- reassigning would cause UI flickering on frequent `session.updated` events since it triggers a full rerender
   if not vim.deep_equal(current_session, merged_session) then
+    -- mutate existing `state.active_session` table in place
+    -- reassigning would cause UI flickering on frequent `session.updated` events since it triggers a full rerender
     for key, value in pairs(merged_session) do
       current_session[key] = value
     end


tests/helpers.lua
```diff
@@ -224,15 +224,26 @@ end
 
 function M.get_session_from_events(events, with_session_updates)
   -- renderer needs a valid session id
-  -- find the last session.updated event
+  -- merge session.updated events and use the latest updated session
```
disrupted, line 227: had to change the test helper to match the new runtime behavior since we now merge the `session.updated` events

lua/opencode/ui/renderer.lua
```diff
@@ -839,8 +839,28 @@ function M.on_session_updated(properties)
...
+    return
+  end
+
+  local current_session = state.active_session
+  local revert_changed = not vim.deep_equal(current_session.revert, updated_session.revert)
+  local previous_title = current_session.title
+
+  local merged_session = vim.tbl_deep_extend('force', vim.deepcopy(current_session), updated_session)
+  if not vim.deep_equal(current_session, merged_session) then
+    for key, value in pairs(merged_session) do
```
sudo-tee, line 857: I feel that you should just do a `state.active_session = merged_session`

Doing this might break reactivity of the state.

disrupted: that's what I tried initially but it caused heavy flickering of the UI whenever a `session.updated` event was received (which can be frequent)

disrupted: this is the relevant commit https://github.com/sudo-tee/opencode.nvim/pull/243/changes/d9de04a40d1a4851af0aa2b452ef7b670cbfebf2

sudo-tee: Ok I see what you mean, the timestamps might be different each time.

So in the case it's on purpose. 

Just add a comment before the loop so we know why in the future

disrupted: done! I've added a comment explaining why we mutate instead of reassign

PR #258, opened 2026-02-09
fix(question_window): render multiline question

disrupted:
@sudo-tee I only now discovered the interactive replay commands. man what an amazing debug tool


tests/data/multiple-question-ask-reply-all.json
disrupted, line 1: what do you think about formatting these input files? would make it easier to read them and compare the diff

sudo-tee:
> @sudo-tee I only now discovered the interactive replay commands. man what an amazing debug tool

You can thank @cameronr for this amazing tool :)

tests/data/multiple-question-ask-reply-all.json
sudo-tee: Yes this would be useful indeed

sudo-tee:
Thanks for the PR

Looks good to me

PR #260, opened 2026-02-10
fix(input_window): disable dynamic height if min_height == max_height

### LINKED ISSUE #253: Dynamic height (09bc7d4) breaks edgy.nvim compatibility and causes hang on quit
Commit https://github.com/sudo-tee/opencode.nvim/commit/09bc7d46f065634af7fc879b6d916fa4353b185e makes `opencode.nvim` unusable when using `edgy.nvim` for window management. The cursor jumps to the beginning of the line on every character typed, making it impossible to write prompts.
Additionally, after this commit, Neovim hangs for some time on quit even when not using `edgy.nvim` at all.

I used this snippet to "fix" edgy.nvim issue for me:
```lua
config = function(_, opts)
  require('opencode').setup(opts)

  local input_window = require 'opencode.ui.input_window'
  input_window.schedule_resize = function() end
end,
```

But I am not sure about the reason of the hanging.

disrupted:
fix #253

### DIFF
--- a/tests/unit/zoom_spec.lua
+++ b/tests/unit/zoom_spec.lua
@@ -127,7 +127,43 @@ describe('ui zoom state', function()
       assert.equals(80, state.pre_zoom_width)
     end)
 
-    it('does not call update_dimensions for fixed input height (dynamic height disabled)', function()
+    it('calls update_dimensions when dynamic height is enabled', function()
+      config.setup({
+        ui = {
+          input = {
+            min_height = 0.1,
+            max_height = 0.25,
+          },
+        },
+      })
+
+      package.loaded['opencode.ui.input_window'] = nil
+      local dynamic_height_input_window = require('opencode.ui.input_window')
+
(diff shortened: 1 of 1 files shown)


sudo-tee:
This looks like a reasonable fix.

I know there is none right now, but we should add a little bit of testing around the dynamic resize so we don't end up breaking it in the future

disrupted:
> This looks like a reasonable fix.
> 
> I know there is none right now, but we should add a little bit of testing around the dynamic resize so we don't end up breaking it in the future

I agree! which type of test would you prefer for it? replay/unit?

sudo-tee:
I don't think you can replay the input behavior. Just a quick unit test.

disrupted:
@sudo-tee done! added 2 basic unit tests for dynamic height enabled/disabled

sudo-tee:
Looks good to me

Thanks for the PR and the test

PR #263, opened 2026-02-12
fix(config_file): respect hidden agents and avoid duplicates

hermandavid:
This PR implements support for [hidden](https://opencode.ai/docs/agents/#hidden) option in OpenCode subagents config.
Also, it avoids duplicate completion options when default agents are defined in OpenCode config.


sudo-tee:
Thanks for the PR 

It looks good to me.

PR #265, opened 2026-02-12
fix(ui): use cursor position for scroll-intent tracking

jensenojs:
- output_window: is_at_bottom() checks cursor instead of viewport (viewport almost always at bottom in streaming chat)
- CursorMoved autocmd replaces WinScrolled for both windows
- state: add cursor persistence functions (save/get/set)
- output_window: add buffer_valid() guard for buffer-only operations

### DIFF
--- a/lua/opencode/state.lua
+++ b/lua/opencode/state.lua
@@ -196,6 +196,68 @@ function M.is_running()
   return M.job_count > 0
 end
 
+---@param pos any
+---@return integer[]|nil
+local function normalize_cursor(pos)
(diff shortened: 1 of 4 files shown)


jensenojs:
for 

> I find one issue where if you move the cursor in the outpout window withoud scrolling and close/reopen, the cursor will be set at the bottom instead of restoring to the last position.

- https://github.com/sudo-tee/opencode.nvim/pull/229#issuecomment-3872530017

sudo-tee:
Thanks,

This one make sense and is pretty straight forward

PR #266, opened 2026-02-12
feat(ui): add reference_picker_center_on_jump config

jensenojs:
- Default false : center viewport when jumping to code references.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ require('opencode').setup({
     display_context_size = true, -- Display context size in the footer
     display_cost = true, -- Display cost in the footer
     window_highlight = 'Normal:OpencodeBackground,FloatBorder:OpencodeBorder', -- Highlight group for the opencode window
+    reference_picker_center_on_jump = false, -- Center viewport (zz) when jumping to a code reference (default: false)
     icons = {
       preset = 'nerdfonts', -- 'nerdfonts' | 'text'. Choose UI icon style (default: 'nerdfonts')
       overrides = {}, -- Optional per-key overrides, see section below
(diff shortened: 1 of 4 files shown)


jensenojs:
> at 0:13 in the video it looks like the cursor jumped up a little when re-opening the output buffer. do you know why that's happening?

- https://github.com/sudo-tee/opencode.nvim/pull/229#issuecomment-3860302533

@disrupted  what do you think?

sudo-tee:
@jensenojs 

I am not sure about this one, since the reference picker is just opening the the file, the cursor should be where it was the last time ?

I feel that this kind of personal preferences would be best addressed with an autocmd in the user config ?

Was there a specific issue with how the reference picker was opening the files ?

disrupted:
> > at 0:13 in the video it looks like the cursor jumped up a little when re-opening the output buffer. do you know why that's happening?
> 
> * [Currently splitting implementation via new PRs - feat(ui): preserve buffers on toggle #229 (comment)](https://github.com/sudo-tee/opencode.nvim/pull/229#issuecomment-3860302533)
> 
> @disrupted what do you think?

commented on the wrong PR first time 🤦🏻‍♂️

yeah, I have to agree with @sudo-tee here. I would also prefer if we simply restore the cursor position as-is. it shouldn't require any special followup handling such as centering the viewport

jensenojs:
Sorry, I got confused, What I originally wanted to fix was this problem. 

> > at 0:13 in the video it looks like the cursor jumped up a little when re-opening the output buffer. do you know why that's happening?在视频的 0:13 处，重新打开 output buffer 时光标似乎稍微向上跳了一下。你知道这是为什么吗？
> 

But this change has nothing to do with it.. I'll close this

PR #269, opened 2026-02-12
refactor(ui): use locale-aware timestamps & improve duration rendering

disrupted:
- use locale time formatting so 12h/24h follows env settings (previously always AM/PM regardless of locale)
- right-aligned time for a cleaner message header
- show reasoning and tool usage duration in seconds (rather than `{start time} – {end time}`)

UI changes


disrupted:
@sudo-tee what do you think of this? the timestamps should now work better for users who prefer 24h time instead of 12h AM/PM format. I also tried improving the UX around the durations displayed for reasoning and tool usage.

I would like to followup with another PR after this to live track the duration. during reasoning or tool use the seconds counter could automatically increment (1s, 2s, 3s...). Later I would also like to make some improvements around subagents (which ones are finished, which ones are still running) and the global loading indicator ("Thinking...") to hide it while the execution is halted (e.g. permission prompt).

sudo-tee:
Thi is great , I think it does look beter with the right alignment :)

The counter is also a great idea.

Thanks a lot

PR #273, opened 2026-02-13
fix(ui): avoid forced scroll-to-bottom while reading older output

disrupted:
This fixes a regression where incoming events could jump the output to the bottom despite config option `ui.output.always_scroll_to_bottom = false` (most likely introduced in [22ee4f0](https://github.com/sudo-tee/opencode.nvim/commit/22ee4f0f76e7b7de60a57f9a524893e50f0edbc4))
`scroll_to_bottom` now decides based on current cursor position vs previous line count instead of stale viewport flag state

### DIFF
--- a/tests/unit/cursor_tracking_spec.lua
+++ b/tests/unit/cursor_tracking_spec.lua
@@ -185,3 +185,57 @@ describe('output_window.is_at_bottom', function()
     assert.is_true(output_window.is_at_bottom(win))
   end)
 end)
+
+describe('renderer.scroll_to_bottom', function()
+  local renderer = require('opencode.ui.renderer')
+  local output_window = require('opencode.ui.output_window')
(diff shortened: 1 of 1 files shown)


lua/opencode/ui/renderer.lua
```diff
@@ -338,10 +344,6 @@ function M.scroll_to_bottom(force)
     vim.api.nvim_win_call(state.windows.output_win, function()
       vim.cmd('normal! zb')
     end)
-    output_window.viewport_at_bottom = true
-  else
-    -- User has scrolled up, don't scroll
-    output_window.viewport_at_bottom = false
```
disrupted, line 344: this appears to be leftover from previous implementation (dead code)

sudo-tee:
Great thanks for the fix :)

PR #274, opened 2026-02-13
fix(input): prevent cursor jumps on context completion trigger

disrupted:
when inserting the context completion trigger (`#`) the cursor could jump to another location because it tried restoring a formerly saved cursor position. this is likely a regression of [`https://github.com/sudo-tee/opencode.nvim/commit/22ee4f0f76e7b7de60a57f9a524893e50f0edbc4`](https://github.com/sudo-tee/opencode.nvim/commit/22ee4f0f76e7b7de60a57f9a524893e50f0edbc4)

I also added a safeguard so that we never attempt to restore a saved cursor location when the input window is already focused

### DIFF
--- a/lua/opencode/api.lua
+++ b/lua/opencode/api.lua
@@ -364,7 +364,7 @@ end
 
 function M.context_items()
   local char = config.get_key_for_function('input_window', 'context_items')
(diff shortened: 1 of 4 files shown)


sudo-tee:
Another great fix.

Thanks

PR #275, opened 2026-02-13
feat(input_window): keep input visible in new sessions with auto_hide enabled

disrupted:
Prevent auto-hiding the input window when the output window is empty (new session).
This improves UX by making it clear where users should start typing when beginning a conversation. Otherwise it could be confusing when the whole screen is basically blank with nowhere to start typing


sudo-tee:
This is a great improvement 

Thanks for the PR

PR #276, opened 2026-02-14
fix(dialog): respect auto_hide config when dismissing dialogs

disrupted:
Fixes input window incorrectly appearing after dismissing question/permission dialogs despite `config.ui.input.auto_hide=true`


tests/unit/dialog_spec.lua
disrupted, line 1: wasn't sure if the tests belong in `input_window_spec` instead but I guess it should follow the source file 😄

sudo-tee: More tests never hurts. 😂

sudo-tee (approved):

PR #277, opened 2026-02-14
feat(renderer): scroll to bottom on submit

disrupted:
continuing with the small UX polish :)


sudo-tee:
This is actually a really good improvement 

Thanks

PR #278, opened 2026-02-14
fix(event_manager): prevent duplicate listeners on restart

jensenojs:
EventManager:start() was registering anonymous state.opencode_server listeners that couldn't be unsubscribed. On restart (e.g., during UI toggle), multiple listeners would accumulate, causing duplicate event handling.

Store listener reference to enable proper unsubscribe/resubscribe cycle.


sudo-tee:
Thanks for the PR

I am away today. I will review this and test tomorrow

sudo-tee:
@jensenojs 

Thanks for the fix.

Although the bug is here,  in practice the even_manager is only started one due to this check 

image

But if we were to use the stop function in the future we could end up in this situation.

PR #279, opened 2026-02-14
fix(question_window): avoid duplicate 'other' option

disrupted:


tests/data/multiple-question-ask.expected.json
```diff
@@ -490,9 +490,9 @@
     "    2. plenary.nvim - Plenary.nvim provides unit test utilities for Neovim plugins.",
     "    3. other - Other or custom test framework",
     "",
-    "Navigate: `j`/`k` or `↑`/`↓`  Select: `<CR>` or `1-3`  Dismiss: `<Esc>`",
+    "Navigate: `j`/`k` or `↑`/`↓`  Select: `<CR>` or `1-4`  Dismiss: `<Esc>`",
```
disrupted: hm, this still looks off...

disrupted: duplicate 'other' option

for some reason it seems to add a duplicate 'other' option (no. 5)

disrupted: found the issue. I would also like to add a new replay test to cover this before this gets merged

sudo-tee:
@disrupted 

Thanks for the fix :)

PR #280, opened 2026-02-14
fix(renderer): debounce renders, refactor scroll logic, track scrollintent

jensenojs:
- **Debounce renders**: Fix edge case from persist_state benchmarking where rapid session switches trigger redundant renders (harmless in daily use, but a bug nonetheless)
- **Refactor scroll_to_bottom**: nested if-else → condition table, just for code cleanliness  
- **Add WinScrolled tracking**: Capture scroll intent from mouse/trackpad scroll (not just keyboard navigation)


jensenojs:
These accumulated organically while building persist_state—fixing bugs, cleaning up code, adding small enhancements to get the experience right. Even with LLM assistance, splitting them now would be a headache. My apologies for the bundled review. @sudo-tee

sudo-tee:
@jensenojs 

The rendering debounce raises questions.  Simply because  the rendering is made in a way to not allow treesitter to constantly parse the buffer. Treesitter is parsing the buffer as soon as there is "time" between writes.  By adding a debounce/schedule it will degrade the performace in a exponential way as the buffer grows.

It's best to remove this part.

jensenojs:
Removed the debounce logic from `render_full_session` as requested.

just clarify, your concern about treesitter performance (multiple parses) was actually unrelated to this mechanism—`render_full_session` only fetches session data, and buffer writes happen atomically in `_render_full_session_data`, so treesitter only parses once.

However, the deduplication assumed all calls were for the same session. If a user quickly switched sessions (A→B), it would return the previous session's (A) Promise, causing wrong data to render.

I checked my second implementation—now `persist_state` avoids re-fetching sessions when switching states, so removing this deduplication mechanism won't affect subsequent benchmarks (although my previous implementation would have caused memory spikes in this scenario, but that situation itself is uncommon)


Happy New Year by the way
@sudo-tee

sudo-tee:
> Removed the debounce logic from `render_full_session` as requested.
> 
> just clarify, your concern about treesitter performance (multiple parses) was actually unrelated to this mechanism—`render_full_session` only fetches session data, and buffer writes happen atomically in `_render_full_session_data`, so treesitter only parses once.
> 
> However, the deduplication assumed all calls were for the same session. If a user quickly switched sessions (A→B), it would return the previous session's (A) Promise, causing wrong data to render.
> 
> I checked my second implementation—now `persist_state` avoids re-fetching sessions when switching states, so removing this deduplication mechanism won't affect subsequent benchmarks (although my previous implementation would have caused memory spikes in this scenario, but that situation itself is uncommon)
> 
> Happy New Year by the way @sudo-tee

Thanks for the clarification. Either way I prefer not to add complexity when it's not entirely needed.

I will do some tests for the tracking and scrolling behavior

PR #281, opened 2026-02-14
feat(ui): persist_state window/buffer persistence

jensenojs:
When ui.persist_state = true (default), toggle hides windows without deleting buffers and restores them on next toggle, preserving input draft content, scroll/cursor position, focused pane, and input_hidden state.

Three-state window model:
closed → visible → hidden → visible
  ↑___________↓____________↑
   (teardown)   (restore)

Decision engine resolves (status, persist_state, has_display_route, in_tab) → action:

| # | status  | in_tab | persist_state | has_display_route | action         |
|---|---------|--------|---------------|-------------------|----------------|
| 1 | hidden  | any    | true          | any               | restore_hidden |
| 2 | hidden  | any    | false         | any               | close_hidden   |
| 3 | visible | false  | any           | any               | migrate        |
| 4 | visible | true   | any           | true              | close          |
| 5 | visible | true   | false         | any               | close          |
| 6 | visible | true   | true          | false             | hide           |
| 7 | closed  | any    | any           | any               | open           |

Note: Multiple opencode instances across tabs are not currently supported.

### DIFF
--- a/lua/opencode/api.lua
+++ b/lua/opencode/api.lua
@@ -105,39 +105,44 @@ M.toggle = Promise.async(function(new_session)
     state.display_route ~= nil
   )
   local action = decision.action
-
-  local EARLY_RETURN_ACTIONS = {
-    close = M.close,
-    hide = M.hide,
-    close_hidden = ui.drop_hidden_snapshot,
-  }
+  local is_new_session = new_session == true
 
   local function open_windows(restore_hidden)
     local ctx = build_toggle_open_context(restore_hidden == true)
     return core.open({
-      new_session = new_session == true,
+      new_session = is_new_session,
       focus = ctx.focus,
       start_insert = false,
       open_action = ctx.open_action,
     }):await()
   end
 
-  local early_return = EARLY_RETURN_ACTIONS[action]
-  if early_return then
-    return early_return()
-  end
-
-  if action == 'migrate' then
-    -- NOTE: We currently don't support preserving Opencode UI state across tabs.
-    -- If Opencode is visible in a different tab, we tear it down there and recreate it here.
-    if state.windows then
-      ui.teardown_visible_windows(state.windows)
-    end
-
+  local function open_fresh_windows()
     return open_windows(false)
   end
 
-  return open_windows(action == 'restore_hidden')
+  local function restore_hidden_windows()
+    return open_windows(true)
+  end
+
+  local function migrate_windows()
+    if state.windows then
+      ui.teardown_visible_windows(state.windows)
+    end
+    return open_fresh_windows()
+  end
+
+  local action_handlers = {
+    close = M.close,
+    hide = M.hide,
+    close_hidden = ui.drop_hidden_snapshot,
+    migrate = migrate_windows,
+    restore_hidden = restore_hidden_windows,
+    open = open_fresh_windows,
+  }
+
+  local handler = action_handlers[action] or action_handlers.open
+  return handler()
 end)
 
 ---@param new_session boolean?
--- a/lua/opencode/state.lua
+++ b/lua/opencode/state.lua
@@ -268,35 +268,73 @@ function M.is_visible()
   return M.get_window_state().status == 'visible'
 end
 
+---@class OpencodeToggleContext
+---@field status 'closed'|'hidden'|'visible'
+---@field in_tab boolean
+---@field persist_state boolean
+---@field has_display_route boolean
 
-
--- ORDER MATTERS: Rules are evaluated top-to-bottom; first match wins.
--- In particular, the has_display_route rule must precede the persist_state=true/hide rule,
--- otherwise toggling while viewing /help or /commands would hide instead of close.
-local TOGGLE_ACTION = {
-  { when = { status = 'hidden', persist_state = true }, action = 'restore_hidden' },
-  { when = { status = 'hidden', persist_state = false }, action = 'close_hidden' },
-
-  { when = { status = 'visible', in_tab = false }, action = 'migrate' },
-  { when = { status = 'visible', in_tab = true, has_display_route = true }, action = 'close' },
-  { when = { status = 'visible', in_tab = true, persist_state = false }, action = 'close' },
-  { when = { status = 'visible', in_tab = true, persist_state = true, has_display_route = false }, action = 'hide' },
-
-  { when = { status = 'closed' }, action = 'open' },
-}
-
----@param when table
----@param ctx table
----@return boolean
-local function toggle_rule_matches(when, ctx)
-  for key, expected in pairs(when) do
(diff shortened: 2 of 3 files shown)


jensenojs:
~~TODO : A complete video demonstration is needed to show the effects after merging my three pull requests~~

https://github.com/user-attachments/assets/2c3e86b6-b984-4337-a3f1-06f209b25879


图片

benchmark

If you have any suggestions regarding code maintainability, please don't hesitate to point them out, as this is very helpful for my growth and the maintenance of the code.

Nevertheless, I would prefer to complete it further in subsequent PRs. The current split is making me a bit exhausted.

sudo-tee:
@jensenojs 

Is this MR ready to re-review ?

There is some conflicts at the moment against main.

jensenojs:
yes, The core is implementing the persist_state tri-state model (closed/hidden/visible) + decision-driven toggle logic.

Major Changes (4 Files)
| File      | Changes  | Core Content                                                                                                                           |
| --------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------- |
| state.lua | +304/-13 | Added OpencodeHiddenBuffers type, OpencodeToggleDecision type, decision logic resolve_toggle_decision, window state management methods |
| ui.lua    | +276/-29 | Hide/restore window lifecycle, cursor position snapshot/restore, close_windows_impl unifies teardown vs hide handling                  |
| api.lua   | +84/-21  | toggle() completely refactored to decision-table pattern, added hide() API, uses state.resolve_toggle_decision()                       |
| core.lua  | +59/-39  | M.open() supports restore_hidden flow, multiple state.windows → state.is_visible() replacements                                        |

Actually, the `close_windows_impl` function is a bit messy because the specific logic for `hide` and `close` is coupled with the common logic. I'm considering cleaning it up to make it more streamlined later on.

Secondary Changes (Adaptation Adjustments)

| File                              | Changes | Description                                                                   |
| --------------------------------- | ------- | ----------------------------------------------------------------------------- |
| tests/unit/persist_state_spec.lua | +678    | New test file (largest line count)                                            |
| input/output_window.lua           | +12/-4  | Cursor tracking migrated to new API (get_window_cursor + set_cursor_position) |
| footer.lua                        | +5/-2   | Added preserve_buffer parameter support                                       |
| cursor_tracking_spec.lua          | +9/-6   | Tests adapted to new function names                                           |
| input_window.lua                  | +4/-0   | Added cursor position save in _hide() for auto_hide scenario                  |
| 7 other files                     | minor   | README, config, types, autocmds, renderer, etc.                               |


Thanks for your hard work (both you and me). I definitely won't take on such a big PR again. I would rather bother you a few more times. @sudo-tee

lua/opencode/state.lua
```diff
@@ -198,25 +241,128 @@ function M.is_running()
...
+function M.is_visible()
+  return M.get_window_state().status == 'visible'
+end
+
+
+
+-- ORDER MATTERS: Rules are evaluated top-to-bottom; first match wins.
+-- In particular, the has_display_route rule must precede the persist_state=true/hide rule,
+-- otherwise toggling while viewing /help or /commands would hide instead of close.
+local TOGGLE_ACTION = {
```
sudo-tee: I feel that this is almost the same pattern as STATUS_DETECTION ?

Could this be aligned and normalized into an utility ? 

There is also one in the last MR for the scroll called `scroll_conditions`

lua/opencode/ui/ui.lua
```diff
@@ -8,48 +8,276 @@ local topbar = require('opencode.ui.topbar')
...
+function M.close_windows(windows, persist)
+  local should_preserve = persist == true and config.ui.persist_state
+  if should_preserve then
+    return M.hide_visible_windows(windows)
+  end
+  return M.teardown_visible_windows(windows)
+end
+
+-- Finalization strategies for close_windows_impl
+local FINALIZE = {
```
sudo-tee: I feel that this could be move to the  ` M.hide_visible_windows` and `M.teardown_visible_windows`

It's harder to understand this way.

lua/opencode/ui/ui.lua
```diff
@@ -8,48 +8,276 @@ local topbar = require('opencode.ui.topbar')
...
+
+  if M.is_opencode_focused() then M.return_to_last_code_win() end
+  if state.display_route then state.display_route = nil end
 
   pcall(vim.api.nvim_del_augroup_by_name, 'OpencodeResize')
   pcall(vim.api.nvim_del_augroup_by_name, 'OpencodeWindows')
+  pcall(vim.api.nvim_del_augroup_by_name, 'OpencodeFooterResize')
+
+  topbar.close()
+  if should_preserve then
```
sudo-tee: This also could be move to the `M.hide_visible_windows` and `M.teardown_visible_windows`  respectively

lua/opencode/api.lua
```diff
@@ -48,40 +48,96 @@ function M.close()
...
+    }):await()
+  end
+
+  local early_return = EARLY_RETURN_ACTIONS[action]
+  if early_return then
+    return early_return()
   end
 
-  if state.windows == nil or not are_windows_in_current_tab() then
+  if action == 'migrate' then
```
sudo-tee: If you move this in a fnction you dont need the EARLY_RETURN anymore

It mecome more like a switch case 

```lua
local function switch(value, cases)
  local case = cases[value] or cases.default
  if case then
    return case()
  end
end
```
and use it like so in this function 

```lua
local result = switch(action, {
 close = M.close,
    hide = M.hide,
    close_hidden = ui.drop_hidden_snapshot,
    migrate = M.migrate_windows
   default = function() return  open_windows(false) end,
})

```

sudo-tee (commented):
I tested a bit and it works well, I like it overall

I left a couple of comments regarding the structure of the code as this is quite complex.

I feel like this should be simplified a little bit before merging.

Thanks a lot for your efforts on this. 

It's almost there :)

lua/opencode/state.lua
jensenojs: I aligned this to the same ordered-rule pattern and introduced a shared `first_matching_rule(...)` helper.

lua/opencode/api.lua
jensenojs: Agreed. I replaced the `EARLY_RETURN` + special-case branching with a single `action_handlers` dispatch table (`close`, `hide`, `close_hidden`, `migrate`, `restore_hidden`, `open`) plus a default fallback.

lua/opencode/ui/ui.lua
jensenojs: done

sudo-tee:
@jensenojs 

Thanks for addressing the comments I made. 

It definitely is easier to understand this way.

I will do a final testing this weekend and it should be good to go

PR #282, opened 2026-02-16
fix: snacks model picker order

joe-p:
This PR ensures that favorites appear at the top of the picker when using snacks. 

Tested with my local config and it works as expected.

Disclaimer: This PR was mostly done by Codex. If there's a more preferable approach just let me know!


sudo-tee:
@joe-p  Thanks this is the fix I didn't know I needed :)

PR #283, opened 2026-02-17
fix(picker): use index-based item matching in fzf-lua

hermandavid:
Selecting, deleting, or renaming child sessions in the fzf-lua picker did nothing, the picker just closed silently. 

The issue was in `fn_fzf_index` which matched selected fzf lines back to items using string comparison. This broke for child sessions due to formatting/whitespace mismatches between what fzf returns and what `to_string()` produces.


sudo-tee:
@hermandavid 

Thanks for the fix :)

As I don't use fzf_lua myself, I test it less frequently.

PR #287, opened 2026-02-20
refactor(completion): implement lsp completion

sudo-tee:


sudo-tee:
@disrupted 

This is a whip for universal completion via an in process lsp instead of having to maintain multiple completion adapters.

If you have time I would like your opinion

disrupted:
great idea! I like it a lot. I am gonna leave some comments but overall it works well

lua/opencode/lsp/opencode_completion_ls.lua
```diff
@@ -0,0 +1,180 @@
...
+---@param item CompletionItem
+---@param index integer
+---@return lsp.CompletionItem
+local function to_lsp_item(item, index)
+  local source = require('opencode.ui.completion').get_source_by_name(item.source_name)
+
+  local lsp_item = {
+    label = (supports_kind_icons() and '' or (item.kind_icon .. ' ')) .. item.label,
+    kind = vim.lsp.protocol.CompletionItemKind.Text,
+    kind_icon = supports_kind_icons() and item.kind_icon or nil, -- Only include kind_icon if supported
```
disrupted: does it any harm to set the icon for all providers? imo the language server should not know about specific consumer implementations.

lua/opencode/lsp/opencode_completion_ls.lua
```diff
@@ -0,0 +1,180 @@
...
+      log.error('Error in completion handler: ' .. tostring(err))
+      callback(nil, {})
+    end)
+end
+
+---Create the LSP server configuration
+---@return vim.lsp.ClientConfig
+function M.create_config()
+  return {
+    name = 'opencode_completion_ls',
```
disrupted: I wonder if the name sets a too narrow scope for this LSP. I imagine it could be used to provide other editor features as well in the future, e.g. code action to resolve diagnostic under cursor.

lua/opencode/lsp/opencode_ls.lua
```diff
@@ -0,0 +1,180 @@
...
+    end,
+    root_dir = vim.fn.getcwd(),
+  }
+end
+
+---Start the LSP server for a buffer
+---@param bufnr integer
+---@return integer? client_id
+function M.start(bufnr)
+  local config = M.create_config()
```
disrupted, line 223: do we have to reinstantiate this on every buffer? perhaps we can cache it or use `vim.lsp.config`

lua/opencode/ui/completion/commands.lua
```diff
@@ -51,7 +51,7 @@ local command_source = {
           kind_icon = icons.get('command'),
           detail = command.description,
           documentation = command.documentation .. (command.args and '\n\n* This command takes arguments.' or ''),
-          insert_text = command.name,
+          insert_text = command.name:sub(2),
```
disrupted: is this for stripping out the trigger character? I think a comment would be helpful

lua/opencode/ui/completion.lua
```diff
@@ -55,90 +17,122 @@ function M.setup()
...
+  local preferred = config.preferred_completion or config.preferred_completion_engine
+  if preferred and preferred ~= 'vim_complete' then
+    return true
   end
-end
 
-function M.is_visible()
-  if M._current_engine and M._current_engine.is_visible then
-    return M._current_engine:is_visible()
+  local known_engines = {
```
disrupted: kinda related to my other comment. imo it would be better if we can simply provide the LSP without needing to know which specific completion plugin is installed. also we might be unaware of other completion engines that people might use

sudo-tee: This one is because the default Nvim completion does not auto trigger. If I set this for all providers It will show 2 completions at the same time. I also hate this . I'm opened for idea's

lua/opencode/lsp/opencode_completion_ls.lua
sudo-tee: @disrupted  I decided to rely on config instead for this.

now it will check for the option `ui.completion.supports_kind_icons` instead, with a note to enable it for blink.

lua/opencode/lsp/opencode_completion_ls.lua
sudo-tee: Great idea, I renamed it to opencode_ls

lua/opencode/ui/completion.lua
sudo-tee: @disrupted 

I decided to use a config for this. `ui.completion.use_native_completion` for this, I think it's a bit better than trying to find a completion plugin.

sudo-tee:
@disrupted 

I will put this on the side for now,

There is to many issues still, as nvim_cmp does not use CompletionDone, so I never know when a completion is completed.

Also I cannot reliably know when the menu is closed as each provider creates his own menu (used for shortcuts)

Native completion menu is not displaying subsequent search results for files.

So in the end being agnostic results in more hacks then adding support for different engines :/

If you have ideas on how to resolve theses issues don't hesitate.

disrupted:
@sudo-tee no worries. I do believe leveraging LSP would be beneficial, but perhaps a generic completion provider is one of the more difficult integrations to migrate upfront given how much we depend on it. I don't have any immediate ideas to solve these problems either. perhaps we could start with a more minimal in-process LSP that we can then incrementally add features to and increase the surface area. what do you think of a tracking issue for this to increase visibility for other collaborators?

sudo-tee:
@disrupted 

I was able to use another lsp standard to execute commands when CompletionItem contains a command flag, most client will execute a `workspace/executeCommand`, so having this and `CompledionDonePre` it covers most of the completion provider.

For the visibility I de cided to go for engine specific api, it's not ideal but it's better from an UX perspective.

I prefer having one check for engine on the visibility of the menu than multiple completion engine implementation.

The long term benefit of having an lsp is worth a couple of little hacks lit this


I will do some more testing but this is almost ready

disrupted:
@sudo-tee very cool! I am gonna take this for a spin tomorrow and report back in case I notice anything odd

disrupted:
@sudo-tee accepting the completion result seems to be broken for me, at least with blink.cmp. For example I can no longer toggle diagnostics context on and off. If you need more debug info or a small video I am happy to provide it. just out of curiosity, which completion engine are you using?

sudo-tee:
@disrupted 
Thanks for the comments. I am using blink but I didn't notice the issue at first because the first time you accept a completion it works fine. I will have a look

EDIT:
I found the issue, there was a conflict between CompletionDone autocmd and executeCommand the handled flag was handled to early

disrupted:
@sudo-tee now it's working great 🎉

sei-dwtompkins:
Hey @sudo-tee, I'm facing the same issue with completions (via `Tab`) seemingly broken, but I'm using `nvim-cmp`. I still see the completions menu, but hitting `Tab` does not select the completion. Instead it will insert a literal `Tab` into the input buffer.

Config:

```lua
input_window = {
  ['<tab>'] = { 'switch_mode', mode = {'n'}}
}
```

sudo-tee:
@sei-dwtompkins 

Can you open a dedicated issue for this. I will try to have a look at it quickly ?

sei-dwtompkins:
Yes, thanks

Issue #288, opened 2026-02-20
Streamed output only shows once complete


Defuse4045:
Hello, despite using the default options, whenever I prompt an LLM whether it's a proprietary one through an API or a local one exposed via llama.cpp and accessed via opencode, this plugin does not show me the streamed output of the LLM which you typically see in the opencode TUI for example. It only displays the entire output after the model has finished its response (I don't see the model's individual tokens appear one by one, only the final output)
Even when you query a thinking model, it shows "Reasoning", but the "toggle reasoning output" does not work, it only toggles it at the very end once the model is done with all its replies and all of its tool calls.
Is this configurable? Or is this intended? Thanks!

jensenojs:
This appears to be a regression problem.

disrupted:
> This appears to be a regression problem.

since the replay tests on our side show no difference it might be an upstream change in the way the opencode server sends the events

edit: looks like it sends incremental `message.part.delta` events now. previously the full message text would be present in `message.part.updated`

sudo-tee:
@Defuse4045 

The fix has been merged. Thank to @disrupted . 

Le us know if this resolved your issue

Defuse4045:
Okay it works now, thanks @disrupted and @sudo-tee!

PR #290, opened 2026-02-21
fix(event_manager): normalize message.part.delta streams

disrupted:
related to https://github.com/sudo-tee/opencode.nvim/issues/288#issuecomment-3938605296


sudo-tee:
Thanks for the quick response on this issue.

PR #291, opened 2026-02-21
fix: only handle global or instance tab-scoped DirChanged events

disrupted:
this fixes an issue where the plugin would react to unrelated `DirChanged` events such as window-scoped (e.g. Neogit `:lcd`).
I believe we should only react to global cwd changes or if it concerns the tabpage where Opencode is used.


sudo-tee:
Thanks. I did t thought of that 

Great fix again

Issue #289, opened 2026-02-21
[Bug/Feature] nvim plugin doesn't auto-sync when TUI is already running


jensenojs:
When I start `opencode` in the terminal (TUI) first, then open a file in nvim with opencode.nvim, the two clients don't automatically sync messages between each other.

It's a bit embarrassing to say, but I'm currently using opencode.nvim more when I'm debugging, because switching to opencode tui is just as convenient as opening opencode.nvim. It's probably also related to my current development workflow. I do more reading in the tui, and naturally, I also do more typing in the tui.

Alright, before persist_state PR merges, toggle can still auto-refresh in a way. Because it resubscribes to all messages... (lol)

The ideal scenario would be dual-end synchronization. Whether I type in opencode tui or opencode.nvim, the messages should automatically sync when outputted—that would be perfect... It's important to maintain flexibility, as this feature is something I really want... 

So I actually made some simple attempts earlier and managed to achieve synchronization between both ends when starting with opencode.nvim + tui attach. However, it seems like fully implementing this feature might require some modifications to the upstream code.

- https://github.com/anomalyco/opencode/issues/216

jensenojs:
Necessary conditions for bidirectional real-time synchronization: both ends must be connected to the same HTTP server
However, opencode by default starts in internal RPC mode:
- No TCP port listening
- Uses customFetch to go through worker RPC
- nvim plugin unable to detect and attach

The following two exploration documents are in Chinese. But it may be possible to understand their meaning through LLM.

[attach-auto-discovery-v2-60b131.md](https://github.com/user-attachments/files/25457101/attach-auto-discovery-v2-60b131.md)

[tui-nvim-auto-sync-root-cause.md](https://github.com/user-attachments/files/25457106/tui-nvim-auto-sync-root-cause.md)

--

I will try to advance this niche demand myself, I'm just worried about the documentation getting lost.

sudo-tee:
@jensenojs 

The MR for configuring a remote server has been merged. Does this improve the situation for you?

jensenojs:
I've debugged a script. It works. I plan to refine it a bit and contribute it. Perhaps later we can push upstream for some elegant improvements, but until then, this wrapper is also functional. @sudo-tee 

Actually, I think we could start a discussion or maintain a document with some recipes for setting up opencode.nvim. What do you think?

sudo-tee:
@jensenojs 

[...]

PR #295, opened 2026-02-23
Allow configuration of external opencode serve process

sei-dwtompkins:
## Summary
- Adds support for connecting to an external opencode server process instead of spawning one automatically
- Enables developers to connect to containerized or remote opencode server instances

## Changes
- Added `server.url` configuration option
- Modified server initialization to use external URL when provided
- Server process is not spawned when external URL is configured

## Use Cases
- Running opencode server in a Docker container
- Connecting to a remote development server
- Custom server deployment scenarios

## Testing
- Existing tests pass (332 successes)
- Feature tested locally with external server configuration

### DIFF
--- a/lua/opencode/curl.lua
+++ b/lua/opencode/curl.lua
@@ -26,9 +26,6 @@ local function build_curl_args(opts)
   if opts.proxy and opts.proxy ~= '' then
     table.insert(args, '--proxy')
     table.insert(args, opts.proxy)
-  elseif opts.proxy == '' then
-    table.insert(args, '--noproxy')
-    table.insert(args, '*')
   end
 
   if opts.timeout then


nzlov:
maybe auto start docker container by opencode.nvim. bind cwd path. auto rm container when close nvim.

sei-dwtompkins:
As I've been using it, I think that's a better solution. Thanks. Because it's annoying to have to start the container separately. That's a nice feature of this plugin. You might still have to use a custom port/hostname setting though? I don't think it will always be easy to extract a custom port if you're pointing to a container startup script/command.

lua/opencode/api.lua
```diff
@@ -523,6 +523,12 @@ function M.help()
 
   local max_desc_length = math.min(90, vim.api.nvim_win_get_width(state.windows.output_win) - 30)
 
+  -- Ensure M.commands is a table before processing
+  if type(M.commands) ~= 'table' then
+    vim.notify('Error: commands table is not initialized correctly: ' .. type(M.commands), vim.log.levels.ERROR)
```
sudo-tee: I don't think this is relevent to the functionality you are adding

sei-dwtompkins:
But something like `opencode_command = '/Users/myusername/scripts/oc'` would be a nice way to start it up.

lua/opencode/config.lua
```diff
@@ -14,6 +14,10 @@ M.defaults = {
   legacy_commands = true,
   keymap_prefix = '<leader>o',
   opencode_executable = 'opencode',
+  external_server_url = nil,
+  external_server_port = nil,
+  external_server_timeout = 5,
+  external_server_container_cwd = '/app', -- Container path where host cwd is mounted (defaults to '/app')
```
sudo-tee: I would not set this as the default.

If a user want to use a server started locally it should not translate the path

sei-dwtompkins:
And that `oc` script could start the opencode server in the container like:

```bash
PROJ="$(basename "$(pwd)")"
NAME="opencode-${PROJ}"

exec docker run -p 11111:11111 --rm -it \
  --name "$NAME" \
  --add-host=host.docker.internal:host-gateway \
  -e SHELL="/bin/bash" \
  -v "$HOME/.local/state/opencode:/home/node/.local/state/opencode" \
  -v "$HOME/.local/share/opencode:/home/node/.local/share/opencode" \
  -v "$HOME/.config/opencode:/home/node/.config/opencode" \
  -v "$(pwd):/app:rw" \
  opencode "$@"
```

lua/opencode/api_client.lua
```diff
@@ -1,5 +1,50 @@
...
+    return container_cwd .. relative_path
+  end
+  
+  return path
+end
+
+--- URL encode a string for use in query parameters
+--- @param str string The string to encode
+--- @return string encoded_string The URL-encoded string
+local function url_encode(str)
```
sudo-tee: This function is duplicated in server_job.lua

You can move this to the util file

lua/opencode/server_job.lua
```diff
@@ -3,10 +3,100 @@ local curl = require('opencode.curl')
...
+
+--- Try to connect to an external opencode server by checking its health endpoint
+--- @param base_url string The server base URL (host:port)
+--- @param timeout number Timeout in seconds
+--- @return Promise<string|nil> promise Resolves with base URL if successful, rejects if connection fails
+local function try_external_server(base_url, timeout)
+  local promise = Promise.new()
+  
+  -- Try multiple potential health check endpoints
+  local health_endpoints = {
```
sudo-tee: Maybe add this as a config ? 
If it's a config we might need only one ?

lua/opencode/api.lua
```diff
@@ -523,6 +523,12 @@ function M.help()
 
   local max_desc_length = math.min(90, vim.api.nvim_win_get_width(state.windows.output_win) - 30)
 
+  -- Ensure M.commands is a table before processing
```
sudo-tee: I am not afan of LLM comments like this.

I prefer to keep comments only for non abvious things

sudo-tee (changes requested):
Thanks for having a look into this.

I left a couple of comments on the MR

sei-dwtompkins:
Thanks for those notes, I'll try to get some time to cook on this and clean things up.

jensenojs:
Sorry to bother you. But it seems this PR is related to that issue — am I understanding correctly? @sei-dwtompkins (If so, that would be really great)

- https://github.com/sudo-tee/opencode.nvim/issues/289

I'll follow up on the context here more carefully after I wake up

sei-dwtompkins:
Hey @jensenojs, this is addressing a much different problem; but it could present a very specific solution to your issue. However, I think it'd be best not to conflate the two issues.

With this PR, you would be able to:
1. Open Neovim
2. In a separate shell, start opencode server (`opencode serve --hostname 0.0.0.0 --port 11111`)
3. Back in Neovim, run `:Opencode`, connecting to the running opencode server (with `external_server_url=127.0.0.1` and `external_server_port=11111` as new configuration options)
5. In a new shell, run `opencode attach localhost:11111` to attach to the running session from Neovim
  - at this point, everything that happens in Neovim also happens in the running Opencode TUI

I haven't tried it backwards (going from TUI to opencode.nvim), but I expect it would function the same way...

jensenojs:
> Hey @jensenojs, this is addressing a much different problem; but it could present a very specific solution to your issue. However, I think it'd be best not to conflate the two issues.嘿，这解决的是一个非常不同的问题；但它可能会为你当前的问题提供一个非常具体的解决方案。不过，我认为最好不要将这两个问题混为一谈。
> 
> With this PR, you would be able to:通过这个 PR，你将能够：
> 
>     1. Open Neovim  打开 Neovim
> 
>     2. In a separate shell, start opencode server (`opencode serve --hostname 0.0.0.0 --port 11111`)在一个单独的 Shell 中，启动 opencode 服务器（ `opencode serve --hostname 0.0.0.0 --port 11111` ）
> 
>     3. Back in Neovim, run `:Opencode`, connecting to the running opencode server (with `external_server_url=127.0.0.1` and `external_server_port=11111` as new configuration options)回到 Neovim 中，运行 `:Opencode` ，连接到正在运行的 opencode 服务器（使用 `external_server_url=127.0.0.1` 和 `external_server_port=11111` 作为新的配置选项）
> 
>     4. In a new shell, run `opencode attach localhost:11111` to attach to the running session from Neovim在新的 Shell 中，运行 `opencode attach localhost:11111` 以从 Neovim 附加到正在运行的会话
> 
> 
>     * at this point, everything that happens in Neovim also happens in the running Opencode TUI此时，Neovim 中发生的所有事情也会在正在运行的 Opencode TUI 中发生
> 
> 
> I haven't tried it backwards (going from TUI to opencode.nvim), but I expect it would function the same way...我还没有尝试反向操作（从 TUI 到 opencode.nvim），但我预期它的功能是一样的...

Thank you for your explanation! I seem to have tested this scenario. When opencode.nvim is opened first, it exposes a port, and then after the opencode TUI attaches, it can achieve the effect of synchronization between both ends. You can also try it, but if the TUI is opened first, this scenario cannot be achieved.

My ultimate expectation is that regardless of whether tui or opencode.nvim starts first (or exits midway and restarts), it can automatically attach to the same server. This would achieve synchronization between both ends.

jensenojs:
Thank you for your explanation! I seem to have tested this scenario. When opencode.nvim is opened first, it exposes a port, and then after the opencode TUI attaches, it can achieve the effect of synchronization between both ends. You can also try it, but if the TUI is opened first, this scenario cannot be achieved.

My ultimate expectation is that regardless of whether tui or opencode.nvim starts first (or exits midway and restarts), it can automatically attach to the same server. This would achieve synchronization between both ends.

sei-dwtompkins:
@sudo-tee I pushed an update. I think everything's working pretty well. Check out the README.md for example configuration and lmk what you think.

It seems to be working really well locally, and the fallbacks are graceful.

lua/opencode/config.lua
```diff
@@ -14,6 +14,12 @@ M.defaults = {
   legacy_commands = true,
   keymap_prefix = '<leader>o',
   opencode_executable = 'opencode',
+  custom_server_enabled = false,
+  custom_server_url = nil,
+  custom_server_port = nil,
+  custom_server_timeout = 5,
+  custom_server_command = nil,
+  container_cwd = nil,
```
sudo-tee: I would also allow a function there

There is a the wsl usecase where path mapping might be more complex.
#299

I feel that this PR could cover both usecase

sudo-tee:
@sei-dwtompkins  Thanks I will give it a spin today

lua/opencode/config.lua
```diff
@@ -14,6 +14,12 @@ M.defaults = {
   legacy_commands = true,
   keymap_prefix = '<leader>o',
   opencode_executable = 'opencode',
+  custom_server_enabled = false,
```
sudo-tee: I think we should create a server section in the config.

What do you think ?

lua/opencode/server_job.lua
```diff
@@ -133,13 +231,130 @@ function M.stream_api(url, method, body, on_chunk)
...
+    local current_dir = vim.fn.getcwd()
+    local mapped_dir = check_port_mapping(custom_port, current_dir)
+    
+    if mapped_dir then
+      log.warn('try_connect_to_custom_server: port %d is already mapped to directory %s but current directory is %s', 
+        custom_port, mapped_dir, current_dir)
+      
+      vim.notify(
+        string.format(
+          'Warning: OpenCode server at %s (port %d) is already serving a different directory:\n' ..
```
sudo-tee: I am not sure it's a warning, 
There is a valid case to run 1 server for multiple instance.
The plugin already sends the directory option to the server so it should be ok.

Maybe we can add an explanation in the docs about doing this ?

lua/opencode/server_job.lua
```diff
@@ -133,13 +231,130 @@ function M.stream_api(url, method, body, on_chunk)
...
+        vim.log.levels.WARN
+      )
+      M.spawn_local_server(promise)
+    end
+  end)
+end
+
+--- @param promise Promise<OpencodeServer>
+function M.spawn_local_server(promise)
   state.opencode_server = opencode_server.new()
```
sudo-tee, line 288: I would expect to start the local server with the specified port.

The use case might be to spwan 1 server for all nvim instance.

But by doing that we need to make sure no other instance is using it before killing it.

We could register the nvim pid for each port, and kill the process with the last remaining nvim instance.

maybe the port_mapping file could contain the info ?

lua/opencode/server_job.lua
```diff
@@ -133,13 +231,130 @@ function M.stream_api(url, method, body, on_chunk)
...
+    log.warn('try_connect_to_custom_server: failed to connect to custom server: %s', vim.inspect(err))
+    
+    if config.custom_server_command and custom_port and custom_url then
+      log.debug('try_connect_to_custom_server: server not running, executing custom_server_command')
+      vim.notify(
+        string.format('Custom server not found at %s, attempting to start it...', base_url),
+        vim.log.levels.INFO
+      )
+      
+      local ok, result = pcall(config.custom_server_command, custom_port, custom_url)
```
sudo-tee: If we have a custom command we shoud track it's job and pid ?

If we set the custom_port to 'auto' we will end up with multiple opencode process running not being killed when we exit nvim.

This lines up with other comment I made for local server with specified port.

sei-dwtompkins:
I built my own opencode image, but I think the standard anomalyco one should work?
docker pull ghcr.io/anomalyco/opencode:latest

sei-dwtompkins:
Thanks for the feedback, I like all these ideas.

lua/opencode/server_job.lua
sei-dwtompkins: @sudo-tee The issue I was looking at here was this scenario:
1. I've configured `:Opencode` to spawn a containerized opencode server instance, but port is set to 8080
2. I open a project in Neovim and run `:Opencode`
3. I open a new terminal and change into a different project directory, open Neovim and run `:Opencode`
4. Now the latest Neovim instance is connected to the old container, without the current project files mounted— because we can't/shouldn't mount them over the previously mounted project files.

I think you're right if there _isn't_ a container command, then this is probably a valid use case... I'll try to think what the best way might be to check for that without adding too many config directives.

lua/opencode/server_job.lua
sei-dwtompkins: I agree with the PID tracking in the table.

@sudo-tee I wonder if `auto_kill = boolean` or something should be an additional config option?

My thoughts with this were that if you're starting the opencode server process in a container, there's some overhead to starting/stopping the container every time. Right now there's a check to see if the server is already running before spawning (when custom server is enabled). So it also shouldn't spawn additional servers. I could see this being desirable in some use cases?

Also, FYI, I think by mounting the opencode/state volume, the actual Opencode session is not ephemeral, so there shouldn't be any session data loss if configured that way. I'm not 100% certain, but I think that's what I've observed.

```
...
-v ~/.local/state/opencode:/home/node/.local/state/opencode \
...
```

sudo-tee: Yes I think it make sense, to have the auto_kill config, for that purpose.

sei-dwtompkins:
@sudo-tee Some additional updates. I added the `kill_command` for the user to specify how they'd like to cleanup. This is paired with the auto_kill option. If starting a local server, 'auto_kill' will work out of the box— but I think keeping the container cleanup as a custom hook like that is going to make it more flexible for complex setups.

Lmk what you think when you have time.

sei-dwtompkins:
Updated configuration examples in the README.md. I haven't tested the WSL/Windows path translation as I don't have access to a Windows machine currently. So that could use some verification...

sudo-tee:
Thanks 

I think there is something that does not work (EDIT: I had an outdated mapping file, it works now)

I am not able to connect to a running server 

```
[DEBUG Thu Feb 26 14:11:45 2026] /home/francis/Projects/_nvim/opencode.nvim/lua/opencode/log.lua:22: try_custom_server: health check passed
[DEBUG Thu Feb 26 14:11:45 2026] /home/francis/Projects/_nvim/opencode.nvim/lua/opencode/log.lua:22: try_connect_to_custom_server: connected to custom server after starting at http://127.0.0.1:40323 (attempt 1)
[DEBUG Thu Feb 26 14:11:45 2026] /home/francis/Projects/_nvim/opencode.nvim/lua/opencode/log.lua:22: try_connect_to_custom_server: retry 1 failed, will retry again: "...Projects/_nvim/opencode.nvim/lua/opencode/server_job.lua:113: attempt to index local 'mapping' (a string value)"
[DEBUG Thu Feb 26 14:11:45 2026] /home/francis/Projects/_nvim/opencode.nvim/lua/opencode/log.lua:22: try_connect_to_custom_server: scheduling retry 2/3 after 4000ms

```

I also have a question about the spawn_command

If I set a port and url and no spawn_command, I would expect the plugin to spawn a local server with the provided host and port.

sudo-tee:
I did other tests, and still got this warning

```
[opencode.nvim] Note: OpenCode server at http://127.0.0.1:40323 (port 40323) may be configured to serve a different project directory:
Mapped directory: /home/francis/Projects/_nvim/opencode.nvim
Current directory: /home/francis/Projects/ost-achievements

To run isolated servers per project, configure server.port = "auto" in your config.
```

I unserstand your usecase with docker and path mapping, but settting a common server for all nvim instance sould not emit a warning like this.

Use case 1 Docker:

-  server url + port = 'auto'
- spawn command start the docker and set a random port.
- kill command kills the docker process..

Use case 2 Local:

- server url +port = 1234 and no spawn_command
- the plugin start a server at the specified port if it does not exist and does not need path_mapping
- Other nvim instance can connect to the server.
- When last instance leave it kills the process.

sei-dwtompkins:
I verified the first issue, that was a regression, and I agree that it should work the way you described. I will push up a fix.

Let me look at the opencode server API a little and figure out how to do that cleanly. I think right now it will keep the first project directory it started on if we're sharing the server— but I think we'd want to make that the directory it starts in.

sei-dwtompkins:
This is tricky... but I think this provides some saner default behavior across the board. Take a peek and let me know what you think.

lua/opencode/server_job.lua
```diff
@@ -47,22 +47,78 @@ local function save_port_mappings(mappings)
...
+--- @param started_by_nvim boolean
+local function kill_orphaned_server(port, started_by_nvim)
+  if not started_by_nvim then
+    log.debug('clean_stale_pids: port %d server not started by nvim, skipping kill', port)
+    return
+  end
+
+  log.info('clean_stale_pids: killing orphaned server on port %d (no connected clients)', port)
+  
+  local base_url = string.format('http://127.0.0.1:%d', port)
```
sudo-tee: I think we should use the server url and not hardcode it here ?

lua/opencode/opencode_server.lua
```diff
@@ -217,6 +252,78 @@ function OpencodeServer:spawn(opts)
...
+
+--- Attach to an existing opencode server
+--- @param opts OpencodeServerAttachOpts
+--- @return Promise<OpencodeServer>
+function OpencodeServer:attach(opts)
+  local log = require('opencode.log')
+
+  local cmd = {
+    config.opencode_executable,
+    'attach',
```
sudo-tee: The attach method is for starting a a full TUI,  We can not use it this way

lua/opencode/server_job.lua
sei-dwtompkins: Yea, I think you're right... and if localhost/127.0.0.1 is not setup as the default, I think that ought to be the case for that config option. I can double-check.

lua/opencode/opencode_server.lua
sei-dwtompkins: I didn't understand that, shoot. So it's not going to be good because it's starting a separate process and we don't want that overhead?

sei-dwtompkins:
@sudo-tee I think the `attach` command is the only one that provides a `--dir` option, which was why I went for that. I'm struggling to find a way to start in the right project directory when connecting to the server. Example:

```lua
server = {
  url = '127.0.0.1',
  port = 1234
}
```

1. Terminal 1, navigates to project A and opens Neovim, runs `:Opencode`
2. Opencode.nvim opens, starts `opencode serve --port 1234 --hostname 127.0.0.1`
2. Terminal 2, navigates to project B and opens Neovim, runs `:Opencode`
3. Opencode.nvim opens, connects with `opencode attach --dir <pwd> http://127.0.0.1:1234`

EDIT: Is the correct approach to use `opencode run --dir <dir> --attach http://<url>:<port> <prompt>`?
I can rework if that's what I missed...

lua/opencode/server_job.lua
sei-dwtompkins: @sudo-tee I think the problem is we need a specified kill command if it's not a local server. I think I'm checking on the spawn_command rn, but maybe I should be checking on the URL. If it's a local server, it's easy to track and kill, but if there's a spawn_command to start in a container or a remote URL— then we probably _need_ that spawn_command/kill_command spec. Does that make sense, or am I not thinking about that correctly...

sudo-tee:
> @sudo-tee I think the `attach` command is the only one that provides a `--dir` option, which was why I went for that. I'm struggling to find a way to start in the right project directory when connecting to the server. Example:
> 
> ```lua
> server = {
>   url = '127.0.0.1',
>   port = 1234
> }
> ```
> 
> 1. Terminal 1, navigates to project A and opens Neovim, runs `:Opencode`
> 2. Opencode.nvim opens, starts `opencode serve --port 1234 --hostname 127.0.0.1`
> 3. Terminal 2, navigates to project B and opens Neovim, runs `:Opencode`
> 4. Opencode.nvim opens, connects with `opencode attach --dir <pwd> http://127.0.0.1:1234`
> 
> EDIT: Is the correct approach to use `opencode run --dir <dir> --attach http://<url>:<port> <prompt>`? I can rework if that's what I missed...

I will also have a look at it `opencode run ` is just for running a promt from the cli.

Myabe my assumption is wrong but since we are sending the ?directory param for each api calls I was under the impression that it would just work.

sei-dwtompkins:
If you have time to play around I'd appreciate it, feel free to toy with this. I can try to get some time to mess with it too. I think when I was testing that aspect out before `attach`, there was another opencode config layer that held onto the original serve start directory. Or maybe the issue was that the directory change would also change the original session's project directory? I was having a hard time getting that to work.

sei-dwtompkins:
I think your original way of running `opencode serve` was essentially setting `server.port = 'auto'` in this PR? It does not share the original server/port:
image

sudo-tee:
I did play around with it and just doing this for attach worked for me

image

I only tested the attach scenario with `port` and `url` it started the server the 1st time and  then on the next nvim instance it  attached to it with the different directory and and I wans able to work normally on each project and the cwd were correct

sei-dwtompkins:
Maybe [this](https://github.com/anomalyco/opencode/issues/3188#:~:text=Add%20%2D%2Dsession%20flag%20to%20attach%20command%20to%20continue%20session%20via%20already%20running%20headless%20server%C2%A0%233165%3A%20Add%20%2D%2Dsession%20flag%20to%20attach%20command%20to%20continue%20session%20via%20already%20running%20headless%20server%20%2D%20Another%20duplicate%20of%20the%20same%20functionality%20that%20was%20implemented) is what we're looking for?

It looks like it will remain headless if the `--session` flag is provided? Does that look right to you?

sudo-tee:
> I think your original way of running `opencode serve` was essentially setting `server.port = 'auto'` in this PR? It does not share the original server/port: image

Yes this was equivalent to running `auto`  as the opencode server will pick any port when not specified

sei-dwtompkins:
That sounds promising!

sudo-tee:
> Maybe [this](https://github.com/anomalyco/opencode/issues/3188#:~:text=Add%20%2D%2Dsession%20flag%20to%20attach%20command%20to%20continue%20session%20via%20already%20running%20headless%20server%C2%A0%233165%3A%20Add%20%2D%2Dsession%20flag%20to%20attach%20command%20to%20continue%20session%20via%20already%20running%20headless%20server%20%2D%20Another%20duplicate%20of%20the%20same%20functionality%20that%20was%20implemented) is what we're looking for?
> 
> It looks like it will remain headless if the `--session` flag is provided? Does that look right to you?

I don't think we ned to spawn a new process for attaching, we just set the url and the port to an already running server.

Even if the server is started in a different directory, we send the ?directory param for all our server requests, the opencode sersupports this already. No need to overcomplicate this scenario.

sei-dwtompkins:
Lmk if you're going to iterate with that, or if you'd like me to look more closely at that. Unfortunately, I'm a bit swamped today.

sudo-tee:
I can have a look , I don't know if it will be today, but yes I can iterate on the functionality.

sei-dwtompkins:
Thanks, lmk if you use my branch if I need to change any permissions. I'm excited for where this is heading! Very cool stuff. Thanks for your feedback and time.

sudo-tee:
Yes I can use your branch, that would be easier

sei-dwtompkins:
If I do get a chance today I'll try to reproduce your test and see if I can make that work. Don't want to step on your toes though. I am still not sure how to handle the remote/local server situation and how to properly integrate that with the server lifecycle. It makes sense to observe the url config, but I think that requires a spawn_command/kill_command to trigger remote/containerized start/stop— unless, and I don't want to open this can of worms rn, we also provide a config to enable the `--mdns` flag for named local hosts (i.e., http://opencode.local) and `/etc/hosts` customization.

sei-dwtompkins:
I pushed another update and got it working without attach, thanks for that advice. There are still some issues with Opencode server lifecycle. The logic for shutting down an opencode server process doesn't always seem to trigger, so I'm seeing some stale processes from `htop` after exiting all running Neovim instances.

lua/opencode/server_job.lua
sudo-tee: @sei-dwtompkins 

I was also playing with it and I think the port mapping file should keep the server pid as well as the url.

sudo-tee:
@sei-dwtompkins 

I am in the process of extracting the port mapping into his own utility, 

This causes a big conflict at the moment with your branch.

But here is the resulting file if you are interested of integrating it.

```lua
local curl = require('opencode.curl')
local log = require('opencode.log')
local config = require('opencode.config')

--- @class PortMappingEntry
--- @field pid number
--- @field directory string
--- @field mode string

--- @class PortMapping
--- @field directory string
--- @field nvim_pids PortMappingEntry[]
--- @field auto_kill boolean
--- @field started_by_nvim boolean
--- @field url string|nil The URL the opencode server is listening on
--- @field server_pid number|nil The PID of the opencode server process (local servers only)

local M = {}

--- @return string
local function file_path()
  return vim.fn.stdpath('data') .. '/opencode_port_mappings.json'
end

--- @return table<string, PortMapping>
local function load()
  local file = io.open(file_path(), 'r')
  if not file then
    return {}
  end
  local content = file:read('*all')
  file:close()
  local ok, data = pcall(vim.json.decode, content or '')
  return ok and data or {}
end

--- @param mappings table<string, PortMapping>
local function save(mappings)
  local path = file_path()
  local file = io.open(path, 'w')
  if not file then
    log.warn('port_mapping: could not open %s for writing', path)
    return
  end
  file:write(vim.json.encode(mappings))
  file:close()
end

--- @param raw number|PortMappingEntry
--- @param fallback_dir string
--- @return PortMappingEntry
local function normalize(raw, fallback_dir)
  if type(raw) == 'number' then
    return { pid = raw, directory = fallback_dir, mode = 'serve' }
  end
  return raw
end

--- Fire-and-forget graceful shutdown request to a server with no clients.
--- @param port number
local function kill_orphaned_server(port)
  log.info('port_mapping: sending shutdown to orphaned server on port %d', port)
  curl.request({
    url = string.format('http://127.0.0.1:%d/global/shutdown', port),
    method = 'POST',
    timeout = 2000,
    proxy = '',
    callback = function(response)
      local ok = response and response.status >= 200 and response.status < 300
      log.debug('port_mapping: port %d shutdown %s', port, ok and 'succeeded' or 'failed')
    end,
    on_error = function(err)
      log.debug('port_mapping: port %d shutdown error: %s', port, vim.inspect(err))
    end,
  })
  -- We don't forcefully kill — the process may be a Docker container or other
  -- managed process; graceful shutdown is sufficient.
end

--- Purge dead nvim PIDs from every mapping and kill any newly-orphaned servers.
local function clean_stale()
  local mappings = load()
  local changed = false

  for port_key, mapping in pairs(mappings) do
    if not mapping.nvim_pids then
      goto continue
    end

    local alive = {}
    for _, raw in ipairs(mapping.nvim_pids) do
      local entry = normalize(raw, mapping.directory)
      local is_alive = vim.fn.getpid() == entry.pid or vim.uv.kill(entry.pid, 0)
      if is_alive then
        table.insert(alive, entry)
      end
      if type(raw) == 'number' or not is_alive then
        changed = true
      end
    end
    mapping.nvim_pids = alive

    if #alive == 0 then
      local port = tonumber(port_key)
      if port and mapping.started_by_nvim then
        kill_orphaned_server(port)
      end
      log.debug('port_mapping: removing port %s (no connected clients)', port_key)
      mappings[port_key] = nil
      changed = true
    end

    ::continue::
  end

  if changed then
    save(mappings)
  end
end

--- Return the directory a port is already mapped to, or nil when the port is
--- either free or already mapped to current_dir.
--- @param port number
--- @param current_dir string
--- @return string|nil
function M.mapped_directory(port, current_dir)
  clean_stale()
  local mapping = load()[tostring(port)]
  if mapping and mapping.directory and mapping.directory ~= current_dir then
    return mapping.directory
  end
end

--- Return an existing port serving current_dir, or nil.
--- @param current_dir string
--- @return number|nil
function M.find_port_for_directory(current_dir)
  clean_stale()
  for port_key, mapping in pairs(load()) do
    if mapping.directory == current_dir and mapping.nvim_pids and #mapping.nvim_pids > 0 then
      local port = tonumber(port_key)
      if port then
        return port
      end
    end
  end
end

--- Record that this nvim instance is using the given port.
--- @param port number
--- @param directory string
--- @param started_by_nvim boolean
--- @param mode? string 'serve'|'attach'|'custom'
--- @param url? string The URL the server is listening on
--- @param server_pid? number The PID of the server process (local servers only)
function M.register(port, directory, started_by_nvim, mode, url, server_pid)
  mode = mode or 'serve'
  clean_stale()

  local mappings = load()
  local port_key = tostring(port)
  local current_pid = vim.fn.getpid()
  local auto_kill = config.server.auto_kill

  if not mappings[port_key] then
    mappings[port_key] = {
      directory = directory,
      nvim_pids = {},
      auto_kill = auto_kill,
      started_by_nvim = started_by_nvim,
    }
  end

  local mapping = mappings[port_key]
  mapping.nvim_pids = mapping.nvim_pids or {}
  if mapping.auto_kill == nil then
    mapping.auto_kill = auto_kill
  end
  if mapping.started_by_nvim == nil then
    mapping.started_by_nvim = started_by_nvim
  end
  if url then
    mapping.url = url
  end
  if server_pid then
    mapping.server_pid = server_pid
  end

  local pid_exists = false
  local updated = {}
  for _, raw in ipairs(mapping.nvim_pids) do
    local entry = normalize(raw, mapping.directory)
    table.insert(updated, entry)
    if entry.pid == current_pid then
      pid_exists = true
    end
  end
  mapping.nvim_pids = updated

  if not pid_exists then
    table.insert(mapping.nvim_pids, { pid = current_pid, directory = directory, mode = mode })
  end

  save(mappings)
  log.debug(
    'port_mapping.register: port=%d dir=%s pid=%d mode=%s started_by_nvim=%s auto_kill=%s url=%s server_pid=%s',
    port,
    directory,
    current_pid,
    mode,
    tostring(started_by_nvim),
    tostring(auto_kill),
    tostring(url),
    tostring(server_pid)
  )
end

--- Remove this nvim instance from a port's client list.
--- Shuts the server down when it was the last client and auto_kill is set.
--- Also shuts down attach-mode processes unconditionally.
--- @param port number|nil
--- @param server any OpencodeServer instance (state.opencode_server)
function M.unregister(port, server)
  if not port then
    return
  end

  clean_stale()
  local mappings = load()
  local port_key = tostring(port)
  local mapping = mappings[port_key]
  if not mapping then
    return
  end

  local current_pid = vim.fn.getpid()
  local remaining = {}
  for _, raw in ipairs(mapping.nvim_pids or {}) do
    local entry = normalize(raw, mapping.directory)
    if entry.pid ~= current_pid then
      table.insert(remaining, entry)
    end
  end
  mapping.nvim_pids = remaining

  local should_shutdown = #remaining == 0 and mapping.started_by_nvim and mapping.auto_kill

  if server then
    if server.mode == 'attach' or should_shutdown then
      log.debug('port_mapping.unregister: shutting down server for port %d', port)
      server:shutdown()
    end
  end

  if should_shutdown then
    mappings[port_key] = nil
  else
    log.debug('port_mapping.unregister: port=%d still has %d client(s)', port, #remaining)
  end

  save(mappings)
end

--- Return the started_by_nvim flag for a port, or false if unknown.
--- @param port number
--- @return boolean
function M.started_by_nvim(port)
  local mapping = load()[tostring(port)]
  return mapping and mapping.started_by_nvim or false
end

return M

```

I have to leave for the rest of the day so you can safely conftinue on the branch

sei-dwtompkins:
Ok, thanks, I'll take a look. I had actually just ripped out where I was manually setting the port with `port = 'auto'` after realizing that the `opencode serve` command did that naturally. But maybe keeping some logic there might help keep track for cleanup.

sudo-tee:
@sei-dwtompkins 

I pushed a working version with the port mapping refactor i did.

There is probably still issues when we kill the process, but I think the flow is easier to reason about now

I see that calling the `global/shutdown` followed by a uv.sleep adds a significant delay when closing nvim

sei-dwtompkins:
Sorry if I mangled your contribution in the rebase. I tried to keep your work as is.

sei-dwtompkins:
I think the PID cleanup appears to be working with static port and 'auto'.... lmk
I screwed up the kill_command though now. I need to fix that. If I start a docker container, I'd like it to clean that up with the kill_command.

sudo-tee:
@sei-dwtompkins 

I updated the kill sequence for local and remote server it should not kill the owned server if there is still clients.
And even if a server is not owned it should try to kill it if it's the last client.

dekaravanhoc:
Hey,

in #299 I added a runtime config and it works also with containers/remote setups. It provides a lot of startup options and an overwrite for path creation. The Readme contains a base example for remote_url and container setup. You would be able tostart a startupscript with the command. The only important thing is, that the process of the startup needs to log the server url like opencode does, so the connection can be established automatically. The official image in the container from the readme does this automatically. 
Maybe you can check this out.

Best regards

sudo-tee:
@sei-dwtompkins 

I did some cleanup and removed the normalization from the port mapping. Since it's a new feature the file should not exist for users. If it causes issues for you you can delete the port mapping file.

I also rebased main on the branch so be sure tu pull before anything

sei-dwtompkins:
@dekaravanhoc Did you take a look at our code changes here? sudo-tee is right, there is a lot of overlap. Probably because I've been working on this branch, I tend to prefer some of our naming strategies for the config options. I wonder if you'd want to try pulling this down and testing it out? I will try to take a closer look at some of the code you laid out. I haven't tested this branch with WSL, and I don't know some of the limitations you're working with on Windows.

Let me make some comments over there too so as to understand some things.. I'm wondering about the dual path options. Is there some unrelated work related to [kepmaps](https://github.com/sudo-tee/opencode.nvim/pull/299/changes#diff-ee8bd08dddd46a18f516f7b20c8ab4f363b99c1035ff0d9a2acfa2fa0fc7e9f4), though, maybe? 


@sudo-tee I realized that I'd still not taken out the `sleep` calls. I was having a hard time getting a consistent process kill without it. You might know a better solution? I'll read through @dekaravanhoc's PR too and see what method they're using.

sudo-tee:
@sei-dwtompkins  

For local server the kill is pretty reliable with my latests changes.

But I don;t have the setup to test with docker or remote server. Maybe in that case it needs the sleep

sei-dwtompkins:
@sudo-tee It's working well for me. I did have to regenerate the opencode_port_mapping.json. I also had to do a ShaDa file cleanup with Neovim's state? That might just be a weird thing during development...

I added a conditional for the `kill_command` which was previously not working. With that, I'm getting great results for working with a configured port and `spawn_command`. I'm looking at an issue with `'auto'` behavior now because it seems like I'm not getting an opencode server connection with just plain:

```lua
server = {
  url = 'localhost',
  port = 'auto',
  auto_kill = true,
}
```

sei-dwtompkins:
I think that bit I added makes sense... I've not found any outstanding issues with that. I'm liking the behavior across the board.

sudo-tee:
@sei-dwtompkins 

Great I will use the branch for a little bit to see If everything is working.

It's coming along nicely 

@dekaravanhoc If you want to try it with your workflow and let us know if it works it would be vrey nice :)

Thanks a lot

sei-dwtompkins:
@sudo-tee I think I am understanding the need for the dual path mapping that @dekaravanhoc introduced... The server may need a backwards-compatible path-mapping for responses. I propose adding a `reverse_path_map` config? Then I think something like the following would work:

EDIT: Or backwards?

```lua
require('opencode').setup({
  server = {
    url = 'localhost',
    port = 'auto',  -- Random port for project isolation
    
    -- Spawn opencode server inside WSL
    spawn_command = function(port, url)
      local cmd = string.format(
        'wsl.exe -e bash -c "opencode serve --hostname 127.0.0.1 --port %d"',
        port
      )
      print(string.format('[opencode.nvim] Starting WSL server on port %d', port))
      return vim.fn.jobstart(cmd, { detach = 1 })
    end,
    
    -- Kill WSL opencode process
    kill_command = function(port, url)
      print(string.format('[opencode.nvim] Stopping WSL server on port %d', port))
      vim.fn.jobstart('wsl.exe -e pkill -f "opencode serve.*--port ' .. port .. '"')
    end,
    
    -- Windows → WSL path translation (for requests)
    path_map = function(host_path)
      if vim.fn.has('win32') == 1 then
        -- Convert C:\Users\... → /mnt/c/Users/...
        local drive, rest = host_path:match('^([A-Za-z]):(.*)$')
        if drive then
          local wsl_path = '/mnt/' .. drive:lower() .. rest:gsub('\\', '/')
          return wsl_path
        end
      end
      return host_path
    end,
    
    -- WSL → Windows path translation (for responses)
    reverse_path_map = function(server_path)
      -- Convert /mnt/c/Users/... → C:\Users\...
      local drive, rest = server_path:match('^/mnt/([a-z])(.*)$')
      if drive then
        local windows_path = drive:upper() .. ':' .. rest:gsub('/', '\\')
        return windows_path
      end
      return server_path
    end,
    
    auto_kill = true,  -- Kill server when last nvim instance exits
  },
})
```

sei-dwtompkins:
Would you like me to push this code up here? Or put in a separate branch?

sei-dwtompkins:
I think we would need @dekaravanhoc's help testing for certain.

sudo-tee:
> Would you like me to push this code up here? Or put in a separate branch?



> Would you like me to push this code up here? Or put in a separate branch?

I think we can combine both in this branch. It will cause less confusion

sei-dwtompkins:
Update with the reverse_path_map function @sudo-tee 
@dekaravanhoc— Please test when you have the chance. Lmk if that gels with your setup and intentions

sudo-tee:
@sei-dwtompkins 

I will ask copilot for a review on this before we merge. 

I usually like doing this whenever I have a big feature as it sometimes finds issues

lua/opencode/server_job.lua
```diff
@@ -132,31 +170,156 @@ function M.stream_api(url, method, body, on_chunk)
   return curl.request(opts) --[[@as table]]
 end
 
+--- @return number|nil port, or nil if we should spawn local instead
+local function resolve_port()
+  local custom_port = config.server.port or 'auto'
+  if custom_port ~= 'auto' then
+    return custom_port
+  end
+
```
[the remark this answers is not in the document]
sudo-tee: I just removed the comment in the README.

lua/opencode/server_job.lua
```diff
@@ -132,31 +170,156 @@ function M.stream_api(url, method, body, on_chunk)
...
+  if not config.server.spawn_command then
+    return port_mapping.find_any_existing_port()
```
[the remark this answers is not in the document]
sudo-tee: Port will be automatically set by the opencode server itself in this case

lua/opencode/port_mapping.lua
```diff
@@ -0,0 +1,292 @@
...
+        if mapping.server_pid then
+          kill_orphaned_server(port, mapping.server_pid)
+        end
+      end
+    elseif is_last_client then
+      log.debug('port_mapping.unregister: last nvim instance for port %d, killing orphaned server', port)
+      if config.server.kill_command and config.server.auto_kill then
+        server:shutdown()
+      elseif server.job then
+        server:shutdown()
```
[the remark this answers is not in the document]
sudo-tee: I clarified the intent with better named variables here. The behavior is correct

the auto_kill is ment to be used with the kill_command. Local server are expected to be managed and killed normally.

lua/opencode/server_job.lua
```diff
@@ -132,31 +170,156 @@ function M.stream_api(url, method, body, on_chunk)
...
+      )
+      promise:resolve(state.opencode_server)
+    end)
+    :catch(function(err)
+      log.warn('failed to connect to %s: %s', base_url, vim.inspect(err))
+      if config.server.spawn_command and custom_port and custom_url then
+        spawn_and_retry(base_url, custom_port, custom_url, promise, timeout)
+      else
+        M.spawn_local_server(promise, custom_port, custom_url)
+      end
```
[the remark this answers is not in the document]
sudo-tee: This is intended so user can specify hostnames if they have a dns setup like `my.dev.local`  

I added a commit stripping `http(s)://` in case of misconfiguration

lua/opencode/port_mapping.lua
```diff
@@ -0,0 +1,292 @@
...
+      table.insert(remaining, entry)
+    end
+  end
+  mapping.nvim_pids = remaining
+
+  local should_shutdown = #remaining == 0 and mapping.started_by_nvim and mapping.auto_kill
+
+  if server then
+    local is_last_client = #remaining == 0 and mapping.started_by_nvim
+    if server.mode == 'attach' then
```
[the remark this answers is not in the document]
sudo-tee: @sei-dwtompkins 
This was a bug that I fixed, now the server has a proper mode field. It fixed a bug where closing the 'server' before the 'client' was leaving ghost processes

sei-dwtompkins:
I pulled your latest updates and I'm not having any issues. Thanks for the cleanup effort.

lua/opencode/curl.lua
```diff
@@ -26,6 +26,14 @@ local function build_curl_args(opts)
   if opts.proxy and opts.proxy ~= '' then
     table.insert(args, '--proxy')
     table.insert(args, opts.proxy)
+  elseif opts.proxy == '' then
+    table.insert(args, '--noproxy')
```
sudo-tee: @sei-dwtompkins 

I fear this might break some setups for those who have a an auto proxy at work.

Was this mandatory for your setup to work ? If so we might need to expose a config for additional curl params

sei-dwtompkins: Let me try taking it out and see what happens. I seem to recall slapping that in to fix a problem, but it might have been a red-herring for something unrelated.

sei-dwtompkins: I am working around a workplace proxy, but I don't have a proxy config set here...

sei-dwtompkins: Yea, sorry, I think that was something I'd put in while I was debugging and is not required/necessary. I ripped out that bit locally and I'm not seeing any issues with containerized or local server configs.

sudo-tee: Thanks for the response,  I will then remove this part just to be sure.

sei-dwtompkins:
At this stage, I'm also interested in the `OPENCODE_SERVER_PASSWORD` [config option](https://opencode.ai/docs/server/#authentication). I wonder if we could simply start the process with `OPENCODE_SERVER_PASSWORD=$OPENCODE_SERVER_PASSWORD opencode serve`? Or if we should wrap that in a config option in order to return a function, like: `OPENCODE_SERVER_PASSWORD=$(pass opencode) opencode serve`? With a containerized workflow, I don't know how much added security this provides; but, it may be a good option for the local server...

sudo-tee:
> At this stage, I'm also interested in the `OPENCODE_SERVER_PASSWORD` [config option](https://opencode.ai/docs/server/#authentication). I wonder if we could simply start the process with `OPENCODE_SERVER_PASSWORD=$OPENCODE_SERVER_PASSWORD opencode serve`? Or if we should wrap that in a config option in order to return a function, like: `OPENCODE_SERVER_PASSWORD=$(pass opencode) opencode serve`? With a containerized workflow, I don't know how much added security this provides; but, it may be a good option for the local server...

I think this could be an enhacement for later ?

Let get this one merged first ?

sei-dwtompkins:
As I'm using this, I have noticed some odd behavior where I'll get a curl-related error on startup. Afterwards, opening `:Opencode` kills the Neovim session, and the only way to get it working again is to rm the port mapping JSON. It's sporadic, though. I'll see if I can catch it again and figure out what's up...

sudo-tee:
> As I'm using this, I have noticed some odd behavior where I'll get a curl-related error on startup. Afterwards, opening `:Opencode` kills the Neovim session, and the only way to get it working again is to rm the port mapping JSON. It's sporadic, though. I'll see if I can catch it again and figure out what's up...

You are using it with a spawn_command I guess ?

sei-dwtompkins:
Yes, I haven't gotten the error again. I ran a `:checkhealth` and found some other issues with my environment. So I'm still trying to get it to happen again. I think it might be a race condition. I was hitting `<leader>og` quickly after typing `nvim`. Here's the config I've been using:

```lua
  server = {
    url = 'localhost',
    port = 'auto',  -- Random port for project isolation
    -- Path mapping: translate host paths to container paths
    path_map = function(host_path)
      local cwd = vim.fn.getcwd()
      -- Replace host project directory with container mount point
      return host_path:gsub(vim.pesc(cwd), '/app')
    end,
    -- Spawn command: start Docker container with opencode server
    spawn_command = function(port, url)
      local dir_name = string.lower(vim.fn.fnamemodify(vim.fn.getcwd(), ":t"))
      local cwd = vim.fn.getcwd()
      local container_name = string.format('opencode-%s', dir_name)

      -- Check if container is already running
      local check_cmd = string.format('docker ps --filter "name=%s" --format "{{.Names}}"', container_name)
      local handle = io.popen(check_cmd)
      local result = handle:read("*a")
      handle:close()

      if result and result:match(container_name) then
        print(string.format("[opencode.nvim] Container %s is already running, skipping start", container_name))
        return true
      end

      -- First, try to stop any existing container with the same name
      os.execute(string.format('docker stop %s 2>/dev/null || true', container_name))

      local cmd = string.format([[
docker run -d --rm \
--name %s \
-p %d:4096 \
-v ~/.local/state/opencode:/home/node/.local/state/opencode \
-v ~/.local/share/opencode:/home/node/.local/share/opencode \
-v ~/.config/opencode:/home/node/.config/opencode \
-v "%s":/app:rw \
opencode:latest opencode serve --port 4096 --hostname '0.0.0.0']],
        container_name,
        port,
        cwd
      )

      print(string.format("[opencode.nvim] Starting OpenCode container: %s on port %d", container_name, port))
      return os.execute(cmd)
    end,
    -- Kill command: stop Docker container when auto_kill is triggered
    kill_command = function(port, url)
      local dir_name = string.lower(vim.fn.fnamemodify(vim.fn.getcwd(), ":t"))
      local container_name = string.format('opencode-%s', dir_name)

      print(string.format("[opencode.nvim] Stopping OpenCode container: %s", container_name))
      return os.execute(string.format('docker stop %s 2>/dev/null', container_name))
    end,
    auto_kill = true,  -- Enable automatic cleanup when last nvim exits
  },
```

sei-dwtompkins:
Maybe just ignore for now. I have a hunch that it was specific to my environment. I'll update if I happen to break it again. I had just upgraded python and node and hadn't reinstalled the Neovim node/python packages.

sudo-tee:
I will do a last test drive tomorrow and if I find nothing it will be good to merge 
Great job on this 

Thanks

sei-dwtompkins:
Thank you! This is a great Opencode integration for Neovim— lots of respect to OSS maintainers like yourself. I appreciate all your patience and help in this contribution.

sudo-tee:
I think it's ready to go.

If there is major issues, users can open issues

Thanks for your effort

PR #299, opened 2026-02-26
Feat/wsl support

dekaravanhoc:
# Request
As a windows user, that runs opencode in WSL as recommended, this plugin should start the server in wsl in the mounted path.

# Implementation
Adds a `use_wsl` config.
When `use_wsl` is set and nvim runs in windows it starts a server in wsl in the mounted path and connects.
All paths are translated from unix to windows and vice verca, when running from windows with `use_wsl`.

Additionally in an extra commit default descriptions for the whole keymap where added - for WhichKey users etc.
But this can/could be ignored for the focus of the PR.

### DIFF
--- a/lua/opencode/api_client.lua
+++ b/lua/opencode/api_client.lua
@@ -2,6 +2,12 @@ local server_job = require('opencode.server_job')
 local state = require('opencode.state')
 local util = require('opencode.util')
 
+local function encode_query_value(value)
+	return tostring(value):gsub('([^%w%-_%.~])', function(ch)
+		return string.format('%%%02X', string.byte(ch))
+	end)
+end
+
 --- @class OpencodeApiClient
 --- @field base_url string The base URL of the opencode server
 local OpencodeApiClient = {}
@@ -80,7 +86,7 @@ function OpencodeApiClient:_call(endpoint, method, body, query)
 
 		for k, v in pairs(normalized_query) do
 			if v ~= nil then
-				table.insert(params, k .. '=' .. tostring(v))
+				table.insert(params, k .. '=' .. encode_query_value(v))
 			end
 		end
 
@@ -443,7 +449,7 @@ function OpencodeApiClient:subscribe_to_events(directory, on_event)
 	local url = self.base_url .. '/event'
 	directory = util.to_server_path(directory)
 	if directory then
-		url = url .. '?directory=' .. directory
+		url = url .. '?directory=' .. encode_query_value(directory)
 	end
 
 	return server_job.stream_api(url, 'GET', nil, function(chunk)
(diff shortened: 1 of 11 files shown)


sudo-tee:
Thanks for the PR

I am a bit hesitant on this because there is also similar use case like `docker` or `remote server` like #295 

I would prefer adding the necessary options/hooks so you could implement this in your config.

There is already the option ` opencode_executable = 'opencode',` where you could put the wsl.exe command there.

What about simply having a path_mapping option that is a function so you could do your translation logic ? 

This would over both wsl and docker use case instead ?

I am open for ideas but I don't want to hardcode to many workflows in the plugin as I will have to maintain them without having access to the specifics setup to test them properly

dekaravanhoc:
Yep you are right, its a very strict lock on one workflow. The `opencode_executable` did not work with the needed args, as it was only one arg, but the biggest issue is the path handling. I removed the strict path handling with two function in the config and also added a runtime object to get the needed command and args. This should be quite flexible.
Also one remark all of this is basically fully written by AI. I just made the planning, testing etc.

sudo-tee:
Thanks for the changes you did on this,

I like it better this way 

I think there is an overlap with another proposed MR https://github.com/sudo-tee/opencode.nvim/pull/295 

I think it's best to wait on this one to apply what is missing as there is also a `start server command` and `path translating` being implemented there. 

Or may be you can collaborate with the author of the other MR so we can have a solution that works for both.  What do you think ?

dekaravanhoc:
Just added a remote_url option and a pre_start_command option to fully support containers / remote server.

sudo-tee:
@dekaravanhoc 

The other MR has been merged, it should cover most of what is done in this MR.

If there is missing peices don;t hesitate to create another MR to add it.

Let me know if you are ok with closing the PR 

Thanks again  for your effort

dekaravanhoc:
Hey, I checked the merged MR. 
It unfortunalty had several problems for me.
- The opencode_executable is still checked (which you probably dont have. When running docker or wsl. And even if, the version check does not really help, as it is not the same exec)
- The command spawning/needing detached processes is not really great for me. Opening up terminals as I run -lc for the profile. The killing of these process also does not work.
- I ended up just creating a cmd for the executable, so its linked to the nvim process and just closes with it. Using only the path overwrites now.

So its working for me now and I wont add something to it I guess. Probably switching opencode to pwsh in the near future, as windows support gets better and better.
Maybe the checking for the opencode_executable / version etc. should be switched of, if a server/spawn_command is set.

Thanks for you great effort and glad the path conversion could help for the MR. 

This can be closed of course. :)

sudo-tee:
Thanks for confirming. Glad you have a working setup for your workflow.

PR #300, opened 2026-02-27
feat(ui): add vim.ui.select rendering option for questions

khaitranhq:
## Summary

Add support for rendering questions using `vim.ui.select` instead of inline in the output buffer. Users can enable this with the new `ui.questions.use_vim_ui_select` config option.

## Changes

- New configuration option `ui.questions.use_vim_ui_select` (default: `false`)
- Implemented `_show_question_with_vim_ui_select()` to render questions via vim.ui.select
- Updated rendering logic to skip buffer operations when vim.ui.select mode is enabled
- Added documentation to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,9 @@ require('opencode').setup({
       preset = 'nerdfonts', -- 'nerdfonts' | 'text'. Choose UI icon style (default: 'nerdfonts')
       overrides = {}, -- Optional per-key overrides, see section below
     },
+    questions = {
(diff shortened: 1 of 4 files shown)


khaitranhq:
I think we can remove `render_question_display` but I want to hear what others think

sudo-tee:
I am ok with the idea as an option.
It used to be this way before https://github.com/sudo-tee/opencode.nvim/issues/197

But a lot of the questions are multi line and you don't always see the full question.

I will have a look at you PR in the coming days

khaitranhq:
Thank you!

sudo-tee:
Seems to work great.

Thanks

PR #303, opened 2026-03-03
feat(lsp): adjust completion textEdit range to replace typed word

khaitranhq:
## Summary
Fix the LSP completion bug where selecting a completion item would duplicate the typed text (e.g., `@te` + "tests" → `@tetests` instead of `@tests`). The issue was caused by incorrect textEdit range calculation in the LSP completion handler.

## Changes
- **Fixed `to_lsp_item` function** in `lua/opencode/lsp/opencode_ls.lua` to correctly calculate the textEdit range
  - Added `word` parameter to accept the text after the trigger character
  - Calculate correct range: `start_char = col - #word`, `end_char = col`
  - This ensures LSP clients replace the typed text instead of appending to it

- **Updated completion handler** to pass the `word` parameter to `to_lsp_item`

- **Added comprehensive test** in `tests/unit/completion_lsp_spec.lua` that validates the textEdit range for the specific bug scenario

## Testing
- ✅ All 37 completion LSP tests pass (36 existing + 1 new)
- ✅ No regressions in existing completion functionality
- ✅ New test validates the exact scenario from the issue: `@te` completion → `@tests`

## Details
The bug occurred because the textEdit range had both `start` and `end` at the same cursor position, which inserted text instead of replacing it. The fix calculates the correct start position as the position right after the trigger character, allowing the LSP client to properly replace the already-typed word.

### DIFF
--- a/lua/opencode/lsp/opencode_ls.lua
+++ b/lua/opencode/lsp/opencode_ls.lua
@@ -91,13 +91,19 @@ end
 ---Convert opencode CompletionItem to LSP CompletionItem
 ---@param item CompletionItem
 ---@param index integer
+---@param params lsp.CompletionParams
+---@param word string Text after the trigger character to be completed
 ---@return OpencodeLspItem
-local function to_lsp_item(item, index, params)
+local function to_lsp_item(item, index, params, word)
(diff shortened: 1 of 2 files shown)


sudo-tee:
Thanks for the fix.

PR #304, opened 2026-03-03
fix(ui): respect custom user layout for all snacks pickers

disrupted:
custom snacks layout was only applied to some picker variants (e.g. session picker), not others (e.g. models picker)


sudo-tee:
Great thanks for the PR

PR #305, opened 2026-03-03
feat(context): expose public API

disrupted:
this would allow context toggles in keymaps and other user configs

I've been using this one in recent weeks and it's super convenient imo :)

```lua
keymap = {
    input_window = {
        ['<C-a>'] = {
            function()
                local enabled =
                    require('opencode.context').toggle_context('current_file')
                if enabled ~= nil then
                    vim.notify('current file context: ' .. (enabled and 'on' or 'off'))
                end
            end,
            desc = 'Toggle current file context',
            mode = { 'n', 'i' },
        }
    }
}
```

### DIFF
--- a/lua/opencode/context.lua
+++ b/lua/opencode/context.lua
@@ -9,6 +9,26 @@ local BaseContext = require('opencode.context.base_context')
 
 local M = {}
 
+---@type table<OpencodeToggleableContextKey, boolean>
+local toggleable_context_keys = {
+  current_file = true,
+  selection = true,
+  diagnostics = true,
+  cursor_data = true,
+  buffer = true,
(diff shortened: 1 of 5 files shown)


sudo-tee:
This is actually a pretty cool idea. 

I will play around with it,  but I really dig this

sudo-tee:
@disrupted 

I would add the api tho the README.md and maybe a quick section about using it the way you described in the in ther repo.

maybe we could also add the toggle_context in the api file and configure it like so

'toggle_context#current_file' where anything after the # is the param ? But this coule be in another PR

disrupted:
@sudo-tee good idea! I am happy to create a follow-up PR for it

PR #306, opened 2026-03-04
fix(context): keep context state and send behavior in sync

disrupted:
I noticed a few discrepancies between context internal state, UI (context bar/completion menu) and what gets actually sent.

- `current_file` could leak into follow-up prompts even when disabled (stale state now gets cleared)
- context bar/completion state could lag behind real state because some context mutations did not emit `state.context_updated_at`
- typo in context bar subscription (`opencode_focused` vs `is_opencode_focused`)

tbh I did not know about the `current_file` delta behavior and why the context bar sometimes shows it dimmed 😄 I've updated the readme with a short description about this feature

### DIFF
--- a/tests/unit/context_spec.lua
+++ b/tests/unit/context_spec.lua
@@ -154,6 +154,92 @@ describe('format_message', function()
     BaseContext.get_current_selection = original_get_current_selection
     BaseContext.get_current_file_for_selection = original_get_current_file_for_selection
   end)
+
+  it('does not include current_file when disabled even if stale current_file exists', function()
+    local ChatContext = require('opencode.context.chat_context')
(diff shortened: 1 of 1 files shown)


sudo-tee:
@disrupted 

These are great fixes :)

Thanks for this.

sudo-tee:
@disrupted is this still in draft or ready to review/merge?

disrupted:
> @disrupted is this still in draft or ready to review/merge?

I wanna test it a bit more. Haven't had the chance to make sure it's stable. I'll keep it a draft for now and let you know when it's ready

disrupted:
@sudo-tee should be ready. I didn't notice any issues

sudo-tee:
Looks all good to me. Thanks

PR #314, opened 2026-03-10
docs : add community recipes for advanced workflows

jensenojs:
- Add recipe for sharing opencode sessions between TUI and nvim plugin.

see https://github.com/sudo-tee/opencode.nvim/issues/289#issuecomment-4030869229

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,21 @@
+# Opencode.nvim Documentation
+
+## Recipes
+
+Community-contributed configurations and workflows for opencode.nvim.
+
+### Available Recipes
+
+- [Bidirectional TUI/nvim Sync](./recipes/bidirectional-sync/README.md) - Share sessions between TUI and nvim plugin seamlessly
+- [Three-State Layout Toggle](./recipes/three-state-layout/README.md) - Instantly switch between code/split/dialog viewing modes
+
+### Contributing a Recipe
+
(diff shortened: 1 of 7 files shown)


docs/README.md
```diff
@@ -0,0 +1,5 @@
+# Opencode.nvim Documentation
+
+## Recipes
```
sudo-tee, line 3: We should add a link to this docs/ from the main README.

This file could also be named README.md I think

sudo-tee:
Thanks for this.

I will certainly add some receipies in the near future.

I just added a small note

docs/recipes/bidirectional-sync/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+**Server not responding?**
+```bash
+# Check health
+curl http://localhost:4096/global/health
+```
+
+## Acknowledgements
+
+Thanks to @sei-dwtompkins for implementing external server configuration support in [PR #295](https://github.com/sudo-tee/opencode.nvim/pull/295), which enables this workflow.
```
sei-dwtompkins: That's kind of you @jensenojs! But @sudo-tee don't feel the need to clutter up docs. This is a work account anyways, so any attributions might disappear if it's ever removed.

jensenojs:
I am very satisfied with these two recipes, and I am also very much looking forward to yours

sudo-tee:
Thanks, this is great.

I will merge this

PR #315, opened 2026-03-10
feat: persist window width across hide/restore

jensenojs:
Save user-adjusted window width ratio when hiding and restore it when restoring hidden windows.

Let situational width adjustments (wide tables, etc.) persist through toggle cycles while keeping config defaults for fresh sessions.

### DIFF
--- a/lua/opencode/state.lua
+++ b/lua/opencode/state.lua
@@ -58,6 +58,7 @@
 ---@field api_client OpencodeApiClient
 ---@field event_manager EventManager|nil
 ---@field pre_zoom_width integer|nil
+---@field last_window_width_ratio number|nil
 ---@field required_version string
 ---@field opencode_cli_version string|nil
 ---@field current_cwd string|nil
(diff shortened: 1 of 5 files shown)


jensenojs:
Sometimes the default width display may not be wide enough (for example, we find that opencode has just output a relatively wide table). In such cases, my instinctive action is to make temporary fine adjustments using a touchpad or similar device. During this temporary adjustment, I do not want the default settings to override it.

My local testing behavior is acceptable so far, but I've only conducted one round of vibe review, and I need to take a closer look myself.

But I have to go to sleep now. If you have time and no issues (or if you have any issues), feel free to make changes directly

jensenojs:
before : 

https://github.com/user-attachments/assets/60299feb-e962-405a-b594-6b8ed1921e04


after : 

https://github.com/user-attachments/assets/f462ae39-6ebc-46e8-8fa9-4acf513a975a

jensenojs:
Do I need to add test cases? @sudo-tee

sudo-tee:
@jensenojs  I would indeed like it if you have time to add tests for this :)

sudo-tee:
Thanks for the fix
Works great :)

PR #319, opened 2026-03-12
fix(core): resolve model via switch_to_mode in ensure_current_mode

hermandavid:
Fix `ensure_current_mode` to delegate to `switch_to_mode` so agent-specific model overrides are applied during startup


sudo-tee:
Thanks for the PR.

This makes sense to me

Issue #318, opened 2026-03-12
Feature : Better diff view


jensenojs:
i also want this

- https://github.com/sudo-tee/opencode.nvim/issues/43#issuecomment-3866110525

sth like

- https://github.com/backnotprop/plannotator
- https://github.com/agavra/tuicr

We can look into this together; I want our code to have sufficient flexibility and maintainability. 

The core code itself should contain as little code as possible, offloading reasonable complexity to our own nvim configuration.

sudo-tee:
It is indeed an interesting concept.

However this can add a lot of complexity. 

There is already a possibility to view the diffs in a session, which uses the snapshots feature of opencode. But it's a but clunky at the moment, and I don't use it that much.

For the planning plugin, it's also an interesting idea, I remember seeing a video by the `opencode` creator on a special flag with a similar workflow for plans, but I haven't checked if it was ever implemented properly. 

https://x.com/thdxr/status/2011204789607284974

jensenojs:
I used plannotator as an example because their recent versions also support code review functionality. I believe that if it's just about the features, an LLM should be able to generate them quite easily. They only need to research how these two frameworks achieved it and migrate it over. We could even consider reusing the diffview plugin.

After all, there's really no difference—it's just about popping up an interface where we can happily comment on things in the nvim style, then package and send it. The code view and plan are the same

But the key here, I think, is how to do it. Opencode has various plugins, and TUI will likely also introduce various plugins soon. As you mentioned in another comment, our core code should be as concise as possible.

---

In my experience, the plan mode of opencode has always been the weaker one compared to claude code and codex... although I am aware of and have enabled the environment variable.

jensenojs:
Agreed, the complexity concern is real. Looking at the codebase, the problem isn't really implementing the diff review UI itself — it's that there's no clean place to put it.

What makes this hard right now: (Sorry , I need AI to help me express the following statement)

- api.lua is a 1593-line catch-all: every command, slash command, and routing logic lives here. Adding diffview integration or plan mode means touching command registration, business logic, and slash command mapping all in the same file — there's no natural boundary to work within.
[...]

PR #337, opened 2026-03-20
refactor(commands): split api into parse/dispatch/handler boundaries

jensenojs:
# Summary

for https://github.com/sudo-tee/opencode.nvim/issues/322

- The main change here is breaking up the command logic that was piled into `api.lua`.

- The execution flow is now split into parse / dispatch / complete / slash layers, with domain behaviors moved into handlers (window/workflow/session/diff/surface/agent/permission). Makes it easier to find and change command behavior without digging through one big file.

- Keymap string actions and `:Opencode ...` now go through the same routing boundary, so the two entry points can't drift independently. Also added a duplicate key fail-fast when aggregating command definitions.

- The dispatch layer has `on_command_before/after/error/finally` hooks wired up but not yet exposed — no config defaults, no subscribers. It's a placeholder for a future event unification pass, not meant to be used yet.


jensenojs:
My modifications here were a bit rushed, but I need to get some sleep (and unfortunately, I had to submit another large PR...). @sudo-tee

sudo-tee:
Thanks for the PR.

I will have a look

jensenojs:
registry part should remove in this pr

sudo-tee:
I was wondering. If the plan is to move to a registry. The core should also use it all the way. I agree that this could be a seperate PR

jensenojs:
That was my local exploratory code, and I still believe some kind of registration mechanism is necessary. It would unify built-in features with those added in the future via recipes or plugins, but it isn't a problem with an immediately obvious, elegant design. I initially wanted to find the answer to this before starting changes on api.lua, but I failed. The time might not be right to solve this problem yet (or perhaps my expectations for it were too high).

sudo-tee:
No worries, take your time, there is no rush

lua/opencode/keymap.lua
```diff
@@ -16,38 +14,56 @@ local function wrap_with_completion_check(key_binding, callback)
...
+  if type(func_name) == 'string' then
+    local command_defs = commands.get_commands()
+    if command_defs[func_name] then
+      return function()
+        local args = func_args and (type(func_args) == 'table' and func_args or { func_args }) or {}
+        commands.execute_command_opts({ args = table.concat(vim.list_extend({ func_name }, args), ' '), range = 0 })
+      end
+    end
+
+    vim.notify('Unroutable string keymap action: ' .. func_name, vim.log.levels.ERROR)
```
sudo-tee: I would put something like `Cannot find keymap action:` it's easier to read

lua/opencode/types.lua
```diff
@@ -474,6 +563,42 @@
...
+---@field kind_hl? string
+---@field priority? number
+---@field meta? table
+
+---@class OpencodeContextSource
+---@field priority? number
+---@field trigger? fun(context: CompletionContext): boolean|Promise<boolean>
+---@field complete fun(context: CompletionContext): OpencodeContextCandidate[]|Promise<OpencodeContextCandidate[]>
+---@field resolve fun(candidate: OpencodeContextCandidate, item?: CompletionItem): OpencodeContextResolveResult|Promise<OpencodeContextResolveResult>
+
```
sudo-tee: I cannot find where theses types (starting from OpencodeContextResolveEffectType)  are used. Did you forget to remove them

lua/opencode/keymap.lua
```diff
@@ -16,38 +14,56 @@ local function wrap_with_completion_check(key_binding, callback)
...
-          raw_callback(func_args)
-        end
-      end
+      local callback = resolve_callback(func_name, func_args)
 
       local modes = config_entry.mode or default_modes
       local opts = vim.tbl_deep_extend('force', {}, base_opts)
-      opts.desc = config_entry.desc or cmds[func_name] and cmds[func_name].desc
+      opts.desc = config_entry.desc
+        or type(func_name) == 'string' and command_defs[func_name] and command_defs[func_name].desc
```
sudo-tee: I had errors on this line about invalid desc.

I would write this line like this instead
```lua 
opts.desc = config_entry.desc or vim.tbl_get(command_defs, func_name, 'desc') or ''
```

sudo-tee (commented):
@jensenojs 

I like the direction it's going.

I left a couple of comments and one issue I had on your branch.

jensenojs:
Thanks for the super quick review — I honestly didn’t expect feedback to come in this fast. @sudo-tee

jensenojs:
I haven’t checked in here for a while—how are things going?

By the way, do you use GitHub Copilot? I was wondering if you frequently encounter compression issues in OpenCode. Does it bother you?  @sudo-tee 

--

Sorry, my email notifications often don't arrive on time. I'll try that branch right now.

sudo-tee:
@jensenojs 

I have been more busy the last weeks, and so submerge in the renderer refactor that I totally forgot about this MR.

I do use copilot and I do have frequent compaction on smaller models hovever on sonnet 4.6 and gpt.5.4 it ok, it does not compact as often.

For this PR specifically I will try to do some testing, to get it merged, it's a nice split up.  for future work.

jensenojs:
No rush on this PR, After the render refactor, a few days still needed of observation @sudo-tee 

 I'm just giving you a few small suggestions for that PR, wait a moment

jensenojs:
I believe the splitting of this api.lua is strongly related to the design issue of "Core isn't feature-agnostic." I have already identified some problems in the current design that need further adjustments. I will revise this PR later and may further explain the direction for subsequent refinement.

- https://github.com/sudo-tee/opencode.nvim/issues/318#issuecomment-4204229864

lua/opencode/commands/handlers/session.lua
```diff
@@ -0,0 +1,463 @@
...
+  session = {
+    desc = 'Manage sessions (new/select/child/compact/share/unshare/rename)',
+    completions = session_subcommands,
+    nested_subcommand = { allow_empty = false },
+    execute = function(args)
+      local subcommand = args[1]
+      local action = session_subcommand_actions[subcommand]
+      if not action then
+        invalid_arguments('Invalid session subcommand')
+      end
```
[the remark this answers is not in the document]
jensenojs: Fixed in 5e47626. Updated `session` handler fallback error to include allowed subcommands for non-`:Opencode` entry points:

`Invalid session subcommand. Use: new, select, child, compact, share, unshare, agents_init, rename`

Also added handler-level regression coverage in `tests/unit/commands_handlers_spec.lua` (`returns actionable invalid subcommand errors for agent/session/diff handlers`).

jensenojs:
The current PR has been developed and is ready for you to try out. @sudo-tee 

I have a relatively complete roadmap, and I will discuss it with you elsewhere later. Thank you for your time and effort

---

Actually, the next step is to gradually disassemble core.lua and migrate it to individual modules (you'll notice there are many forwarding operations to core.lua in the handler).

I believe we don't actually need core.lua -- we just need to distinguish between capabilities that require communication with the opencode server to obtain, and the parts where opencode.nvim has the opportunity to showcase its unique strengths. 
 - Moreover, this task doesn't have much complexity and can be completed in phases.

sudo-tee:
@jensenojs Yes I agree with this. 

Your pr is a really good start in that direction. 

I will do a last review/test and we can merge this

PR #353, opened 2026-04-09
feat: restore scroll behavior from working commit

jugarpeupv:
refactor(ui): simplify output window scroll tracking
    
    Restore ed0c07815e867f3b71f4e012de4781cc24fb29e5 scroll behavior
    
    Replace the sticky `_was_at_bottom_by_win` flag with a new approach that
    tracks the previous line count per window. The `is_at_bottom` logic now
    compares the cursor position to the previous and current line counts,
    making auto-scroll behavior more robust and less dependent on viewport
    events. This also removes redundant comments and unused code, and
    ensures scroll tracking state is reset consistently.

@sudo-tee in your main branch the scroll behavior is not working, i mean, no matter how you define the always scroll to bottom setting, that the buffer is always scrolled to bottom

This pr is to restore working scroll behavior from previous commit


lua/opencode/ui/output_window.lua
```diff
@@ -116,18 +107,13 @@ function M.is_at_bottom(win)
     return true
   end
 
-  -- Prefer the sticky flag when it has been set by scroll/WinScrolled events.
-  -- Fall back to a live viewport check on the very first call (flag is nil).
```
jensenojs, line 120: Quick question: does viewport position still drive is_at_bottom() tracking in this change, or is it now cursor-only?

I often pull the viewport away from the bottom using the touchpad, so I want to confirm whether this part will cause a change in behavior.

jensenojs:
Additionally, I cannot reproduce this issue locally (`851e865`), There could be some subtler bug here

jugarpeupv:
@jensenojs i dont know, the breaking code was introduced in this commit 

commit 2f73a21fa665ab4345a87ebbf32558ba07b055f2
Author: Francis Belanger <francis.belanger@gmail.com>
Date:   Tue Apr 7 09:13:55 2026 -0400

    refactor(renderer): decouple events from buffer manipulation  (#340)


in this PR i am just trying to restore the code as it was before this commit


Anyway i think you could checkout my branch and test your usecases

I tested it and now it has the best behavior, which is to behave like a terminal buffer, when the cursorline is at the end of the buffer, and new content is appended, then the cursorline scrolls with the new content, and when the cursorline is not at the end of the buffer, that means the users is reading something, so the cursorline is not scrolled down as new content is rendered

sudo-tee:
I see what is your intent there. But I changed the scroll behavior in my last refactor because it would often stop scrolling even if the cursor didn't move in the output. Most of the time It would stop scrolling before the last message.

So basically what it not working for you ?

Scroll just don't work ?
Always scroll to bottom config is not working ?

jugarpeupv:
@sudo-tee with the refactor code, every time new content is added to the output window, the cursor is placed at the end of the new content, making a scroll effect, and also, if i press "k" several times to move up the cursor when new content is rendered, then cursor keeps going down as new content comes


i think the best behavior is like i said, to make it work like in neovim terminal buffers:

--> cursor is at end of line --> keep moving cursorline to the new appended content
--> cursor is not at the end of line --> means user is reading something --> do not move cursor to the new end of the buffer as new content arrive

sudo-tee:
@jugarpeupv 

I don't have this issue myself on main, it behave exactly how you describe it.

But I will  still merge it if it helps

PR #356, opened 2026-04-11
fix(session): prevent dispatch args leaking as parent_id in select_session

hermandavid:
# Problem

After the commands refactor (#337), the dispatch layer calls all `execute` functions with the signature `execute(args, range)`. The `select_session` keymap alias was directly referencing `M.actions.select_session`, which interprets its first parameter as `parent_id`.

When triggered via keymap with no arguments, dispatch passes `{}` (empty args table) as the first argument. This causes `core.select_session` to filter sessions where `s.parentID == {}`, which never matches — every session is filtered out, and the user always sees "No child sessions found".

The `/sessions` slash command was unaffected because it routes through the `session` command handler, which calls `M.actions.select_session()` without arguments.

## Fix

Wrap the `execute` binding so dispatch args are not forwarded as `parent_id`, matching the pattern already used by `rename_session` on the adjacent line.

### DIFF
--- a/lua/opencode/commands/handlers/session.lua
+++ b/lua/opencode/commands/handlers/session.lua
@@ -441,7 +441,7 @@ M.command_defs = {
   },
   -- action name aliases for keymap compatibility
   open_input_new_session = { desc = 'Open input (new session)', execute = M.actions.open_input_new_session },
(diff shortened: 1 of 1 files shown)


sudo-tee:
Thanks for the fix

PR #357, opened 2026-04-13
fix(ui): disable statuscolumn in input window

vapourismo:
This PR updates the input buffer to clear the `statuscolumn`, ensuring it matches the behaviour of the output buffer.

Without this change, users with a configured `statuscolumn` may see visual artifacts in the status column of the input buffer.


sudo-tee:
Thanks for the quick fix

PR #358, opened 2026-04-15
fix(ui): disable colorcolumn in windows

vapourismo:
Set `colorcolumn=''` for the input and output window. 

Here is an example of how the colour column shows up in the output buffer. I added red arrows to highlight where characters are marked through the `colorcolumn` setting.

Screenshot 2026-04-15 at 16 47 03

With the fix:

image


sudo-tee:
This is great. I never realized that these artefacts were cursor columns. I thought they were from rendering markdown.


Thanks for the fix

PR #359, opened 2026-04-16
fix(session_picker): switch to remaining session when deleting active one

hermandavid:
## Summary

Fix `<C-d>` in the session picker creating a new session when deleting the active one instead of switching to a remaining session.

## Problem

Deleting the active session always spawned a new session, even when other sessions existed. The user would land on an unexpected blank session, and the picker would show the leftover entries plus the freshly-created one. This was also inconsistent with deleting a non-active session, which simply removes the entry.

## Fix

The delete action now switches to the most recently updated remaining session via `core.switch_session`, and only falls back to `core.create_new_session` when no sessions remain.

## Behavior

- Active session deleted with others remaining → switch to most recent remaining.
- Active session deleted when it's the only one → create new (unchanged).
- Non-active session deleted → unchanged.

### DIFF
--- a/lua/opencode/ui/session_picker.lua
+++ b/lua/opencode/ui/session_picker.lua
@@ -56,12 +56,28 @@ function M.pick(sessions, callback)
 
         local sessions_to_delete = type(selected) == 'table' and selected.id == nil and selected or { selected }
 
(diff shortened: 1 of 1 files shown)


sudo-tee:
Thanks this seems like a greast fix.

PR #361, opened 2026-04-20
feat(devkit): add dependency topology scanner script

jensenojs:
During the process of splitting core.lua, the code does not seem to have become clearer.  https://github.com/sudo-tee/opencode.nvim/pull/360

After running some static checks, it was found that after deleting core.lua, the cycle dependency graph did not decrease but instead increased—the largest strongly connected component expanded from 34 modules to 36 modules. Modules such as services.session_runtime took over the central position originally held by core and merged into the main cycle.

#360 is heading in the right direction, but what it does is more like exposing the issues that were originally hidden in core.lua rather than resolving them.

----

We’ve been refactoring without a global view of the dependency graph, any cleanup we did is just hitting a local optimum—we might be fixing one spot while making the overall structure worse.

This PR provides a tool, and I hope it serves as a good start
- Statically scans require dependencies in lua/opencode/**/*.lua to build a graph.
- Detects Strongly Connected Components (SCC, i.e., circular dependencies) and policy violations (cross-layer violations).
- Generate interactive HTML visualizations and agent-friendly JSON output.

Current baseline (upstream/main): 4 SCCs, with the largest cycle containing 34 modules.

Also worth mentioning is that I vibed a topology.jsonc, which roughly aligns with my previous architectural design diagram, targeting a layered architecture (5 layers: entry / dispatch / capabilities / infrastructure / foundation) and specifying which directional dependencies are prohibited (this is a very rough, very rough direction). https://github.com/sudo-tee/opencode.nvim/issues/322#issuecomment-4215707868

After parsing the topology of the call chain, this part of the rules will be further used to evaluate the rationality of the topology itself.

I'm much less familiar with the project than you are, so I need you to evaluate it @sudo-tee . Once you adjust this jsonc, you'll have a tool to check whether the project's current shape matches your expectations and where it diverges. I can then make further project changes guided by that assessment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,7 @@ deps/
 
 # Local Claude settings (keep out of repo)
 .claude/
+
+# Dependency topology tool artifacts
+scripts/dependency-topology/__pycache__/
+*.html
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,3 +11,13 @@
 - **Comments:** Avoid obvious comments that merely restate what the code does. Only add comments when necessary to explain _why_ something is done, not _what_ is being done. Prefer self-explanatory code.
 - **Config:** Centralize in `config.lua`. Use deep merge for user overrides.
 - **Types:** Use Lua annotations (`---@class`, `---@field`, etc.) for public APIs/config.
+
+## Dependency Topology Tool
+
+Use `scripts/dependency-topology/scan_topology.py` to inspect and track architectural layering.
+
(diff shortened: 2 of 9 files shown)


jensenojs:
Here is an example of usage,

iShot_2026-04-20_19 19 17_under10mb_safe

sudo-tee:
@jensenojs 

Thanks for the script I will have a look at it soon. I have a pretty busy week ahead, so it might take a couple of days before I can test it properly

PR #363, opened 2026-04-24
fix(ui): avoid extraneous empty lines for compacted headers

vapourismo:
There is an extraneous empty line in place of a header line when compaction is on.

I am proposing removing this extra line, assuming that it is indeed extraneous and not a visual cue 😄 

## Before

I marked the sections in question with a red line:

Screenshot 2026-04-24 at 13 15 15

## After

Note the gaps have become smaller as the extraneous empty line is removed.

Screenshot 2026-04-24 at 13 19 15


sudo-tee:
Thanks for the PR.

Looks good to me

PR #364, opened 2026-04-27
feat(session): add sibling and parent session navigation

hermandavid:
## Summary

Adds two new session navigation actions for easier browsing of child session hierarchies.

## Problem

When browsing a child session (e.g., session B, child of A), navigating to a sibling session (another child of A) required going back to the root session list, selecting parent A, then listing its children again. There was also no way to jump directly to a parent session.

## Solution

- **`select_sibling_session`**: Opens the session picker filtered to siblings (children of the same parent). Falls back to root sessions with a notification when the active session has no parent.
- **`select_parent_session`**: Directly switches to the parent session. Notifies when there is no parent.
- **`select_child_session` is unchanged** — it still shows children of the current session, preserving grandchild navigation.

## Navigation model

| From | Action | Result |
|---|---|---|
| Root A | `child` | Picker: children of A |
| Child B (parent=A) | `sibling` | Picker: children of A (siblings of B) |
| Child B (parent=A) | `parent` | Switch directly to A |
| Child B (parent=A) | `child` | Picker: children of B (unchanged) |

## New keymaps

| Keymap | Action | Description |
|---|---|---|
| `<leader>oB` | `select_sibling_session` | Select sibling session |
| `<leader>oP` | `select_parent_session` | Go to parent session |


jensenojs:
need it!

sudo-tee:
This is really great.

Thanks for this

PR #365, opened 2026-04-27
fix: restore pending permissions via REST API after session switch

jbro:
## Summary

- Restore pending permission dialogs after a session switch, mirroring the existing `restore_pending_question` pattern in `question_window.lua`
- Previously, switching sessions would auto-deny all pending permissions and discard them — they are now fetched from the server and re-queued

## Changes

**`api_client.lua`** — Added `list_permissions()` method (`GET /permission`) to fetch pending permissions from the server.

**`permission_window.lua`** — Added `restore_pending_permissions(session_id)` which:
1. Fetches all pending permissions via the new API endpoint
2. Filters to those belonging to the active session (including child task sessions)
3. Re-queues matching permissions into the UI, avoiding duplicates

**`renderer.lua`** —
- Calls `restore_pending_permissions()` alongside `restore_pending_question()` in `render_full_session()`
- Removed the old `reset()` logic that auto-denied all pending permissions on session switch

### DIFF
--- a/lua/opencode/api_client.lua
+++ b/lua/opencode/api_client.lua
@@ -304,6 +304,13 @@ function OpencodeApiClient:unrevert_messages(id, directory)
   return self:_call('/session/' .. id .. '/unrevert', 'POST', nil, { directory = directory })
 end
 
+--- List pending permissions
+--- @param directory string|nil Directory path
+--- @return Promise<OpencodePermission[]>
+function OpencodeApiClient:list_permissions(directory)
+  return self:_call('/permission', 'GET', nil, { directory = directory })
+end
+
 --- Respond to a permission request
 --- @param id string Session ID (required)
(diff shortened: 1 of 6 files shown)


sudo-tee:
Thanks for the PR.

Should we check the permission status ? 

I did something similar for questions a while back
If you look at the `restore_pending_question` function  I had to add a check for resolved questions since it would restore answered questions.

jbro:
> Thanks for the PR.
> 
> Should we check the permission status ?
> 
> I did something similar for questions a while back If you look at the `restore_pending_question` function I had to add a check for resolved questions since it would restore answered questions.

Yeah that makes sense, let me take a look and see if I can figure it out 🙂

sudo-tee:
Great thanks for the PR.

Looks good to me

PR #366, opened 2026-04-27
fix: retry instead of spawning local server when auto_kill=false

jbro:
## Summary

- When connecting to an externally managed server (`auto_kill=false`), if the initial health check fails, retry the connection instead of spawning a new local server that would leak as an orphan process
- The spawned Go binary ignores SIGHUP and `VimLeavePre` only unregisters the port without killing the process, so these orphans survive Neovim exit

## Changes

**`server_job.lua`** — Added a new branch in `try_connect_to_custom_server`: when the connection fails and `auto_kill` is `false`, retries the health check (5 attempts with backoff) instead of falling through to `spawn_local_server()`. Resolves the promise on success or rejects with a clear error after retries are exhausted.


sudo-tee:
This seems to be a reasonable fix.

Thanks for the PR

PR #367, opened 2026-04-27
feat(output): add fold support for tool output

### LINKED ISSUE #351: Feature Request: use folds for reasoning and tool outputs toggles
Currently toggles for reasoning and tool outputs just remove/add content from buffers. It would be great if folds were created in their place instead, and toggle controls simply closed/opened them. This would allow the user to more flexibly manage the display and, if necessary, open reasoning/tool ​​output exactly where they need it.

sudo-tee:
This should resolves #351


sudo-tee:
I've been using this for a couple of days and it seems pretty stable. I will merge this

PR #368, opened 2026-04-28
fix: await send_message chain to eliminate flaky user_message_count test

jbro:
## Summary

- Await the `.and_then`/`.catch` chain in `send_message` so the returned promise reflects the full operation lifecycle, not just the setup phase
- Replace flaky `vim.wait(50ms)` polling in tests with deterministic `:wait()` calls on the promise

## Problem

The `increments and decrements user_message_count correctly` test was consistently failing on `macos-latest, v0.11.4` CI ([example on main](https://github.com/sudo-tee/opencode.nvim/actions/runs/25012018331)). `send_message` fired off the `.and_then`/`.catch` chain without awaiting it. Since `create_message` returned an already-resolved promise, the `.and_then` callback was deferred via `vim.schedule_wrap`. The test used `vim.wait(50, ...)` to poll for the count decrement, which was not reliably sufficient on macOS runners.

## Fix

`send_message` is already `Promise.async` (runs in a coroutine), so adding `:await()` at the end of the chain is the correct fix — it makes the function wait for the full send lifecycle before resolving. The tests then use `:wait()` on the returned promise instead of polling.

### DIFF
--- a/lua/opencode/services/messaging.lua
+++ b/lua/opencode/services/messaging.lua
@@ -83,6 +83,7 @@ M.send_message = Promise.async(function(prompt, opts)
       update_sent_message_count(-1)
       session_runtime.cancel():await()
     end)
+    :await()
 end)
 
 ---@param prompt string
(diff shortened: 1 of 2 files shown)


sudo-tee:
This is a great fix, thanks

PR #369, opened 2026-05-02
refactor(session): merge tree navigation actions into direction + interaction parameters

jensenojs:
The three session navigation actions differ only by direction and picker behavior,
but each is a separate function. Adding forward/backward flat navigation would
make this worse.

`navigate_session_tree(direction, interaction, wrap, empty_policy)` replaces
all three, with forward/backward for global next/prev by `time.updated`.

Before                           |After
---------------------------------|------------------------------------------
select_parent_session                 | navigate_session_tree('parent')
select_child_session                    | navigate_session_tree('child', 'picker')
select_sibling_session                 | navigate_session_tree('sibling', 'picker')
—                                | navigate_session_tree('forward', 'direct', true)
—                                | navigate_session_tree('backward', 'direct', true)

New: forward/backward for flat global navigation by time.updated.
Parameters: direction, interaction (direct|picker), wrap, empty_policy.
Implementation: tree_directions config table, async session list lookup.


jensenojs:
thank David Herman !

sudo-tee:
This is a great addition :)

Thanks

PR #370, opened 2026-05-05
feat(session): allow session ID as direction in navigate_session_tree

jensenojs:
When direction is not a known tree direction, treat it as a session ID and switch directly. Task action now passes metadata.sessionId so S key opens the exact child session.


lua/opencode/commands/handlers/session.lua
```diff
@@ -180,6 +180,26 @@ local function compute_target_index(current_idx, total, direction, wrap)
 end
 
 function M.actions.navigate_session_tree(direction, interaction, wrap, empty_policy)
+  -- If direction is not a known navigation direction, treat it as a target session ID.
+  -- This path runs before the tree lookup so invalid inputs get clear feedback.
```
sudo-tee: We dont really need this comment

lua/opencode/commands/handlers/session.lua
```diff
@@ -576,8 +596,12 @@ M.command_defs = {
     end,
   },
   navigate_session_tree = {
-    desc = 'Navigate session tree (parent/child/sibling/forward/backward)',
+    desc = 'Navigate session tree (parent/child/sibling/forward/backward) or switch to a session by ID',
     execute = function(args)
+      -- If args[1] is not a known direction, treat it as a session ID
```
sudo-tee: Same here, the code is easy enough to read

sudo-tee (commented):
Looks good to me, 

Only left 2 minor comments about removing obvious comments in the code

sudo-tee:
Thanks

PR #371, opened 2026-05-05
DRAFT: feat(ui): float ui

Dimfred:
Hi, 
that's a vibed float ui, I'd refactor it before really submitting it.

Question for now would be, would you even want that? Cause if not, no need to refactor for me.

Also would be happy about any suggestion, how you'd like to have it / whether you already see some bugs.

Cheers

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -133,6 +133,18 @@ M.defaults = {
     input_position = 'bottom',
     window_width = 0.40,
     zoom_width = 0.8,
+    float = {
+      width = 0.95,
+      height = 0.9,
+      row = nil,
+      col = nil,
+      border = 'rounded',
+      gap = 1,
+      zindex = 40,
+      opts = {
+        winblend = 0,
+      },
+    },
     picker_width = 100,
     display_model = true,
     display_context_size = true,
--- a/lua/opencode/state/ui.lua
+++ b/lua/opencode/state/ui.lua
@@ -13,7 +13,7 @@ local store = require('opencode.state.store')
 ---@field output_cursor integer[]|nil
 ---@field output_view table|nil
 ---@field focused_window 'input'|'output'|nil
----@field position 'right'|'left'|'current'|nil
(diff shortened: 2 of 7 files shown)


jensenojs:
Can you simply tell us what the final effect looks like? Including
- Effects that users can perceive
- The scope of code refactoring changes and the expected shape

---

I previously wanted to create a feature, but I found that achieving an elegant implementation would require too much refactoring.

Additionally, in a series of small code refactorings we previously conducted, we found that several refactorings did not interact well with each other (although, as a hobby project, it's unrealistic to expect us to be overly serious). However, I still hope that, as much as possible, there will be potential for synergy in future refactoring directions. Perhaps at some point, the project's infrastructure will truly be refined enough to elegantly implement numerous features through composition.

Thank you in advance for your time!

- https://github.com/sudo-tee/opencode.nvim/issues/318#issuecomment-4204229864
- https://github.com/sudo-tee/opencode.nvim/issues/322#issuecomment-4215707868

Dimfred:
> Can you simply tell us what the final effect looks like? Including
> Effects that users can perceive

The effect is simple instead of having a window left / right, you get a toggleable floating window:

image


> The scope of code refactoring changes and the expected shape

Refactoring I mean more what I submitted, so I just wanted to say I'd like to clean it up and really check whether what I vibed fits the current codebase. I am pretty sure it can be simplified.

EDIT:
I am already seeing things which might be related to my patch, i.e. action selection disappearing sometimes, while I am on another tab. So things are still buggy I have to play around here more.

sudo-tee:
Interesting, 

We had a floating UI in the past that we inherited from goose.nvim (original forked repo). But it was not a "popup" style floating rather a floating panel, and the navigation was cumbersome so I removed it. 

But having a floating window like a popup make sense. 

I will review the code ans see how we can adapt it. 

Thanks for the PR

Dimfred:
Just let me know once you know how you'd like to have it.

sudo-tee:
I had a  look and the code is rather clean the way it is.

This is a great contribution.

Thanks

PR #373, opened 2026-05-09
fix(ui): prevent re-applying style='minimal' on float UI

yct21:
Fixes a bug where the float layout re-applies `style = 'minimal'` after `update_dimensions`, resetting window-local options that were explicitly set by `output_window.setup()`.

## Problem

When using `position = "float"`, window options like `signcolumn = 'yes'` and `foldcolumn = '1'` (set by `output_window.setup()`) get reset to `'auto'` and `'0'`. This happens because:

1. `output_window.setup()` sets `signcolumn = 'yes'` and `foldcolumn = '1'`
2. `update_dimensions()` is called immediately after
3. `float_layout.update()` → `nvim_win_set_config()` re-applies `style = 'minimal'` from `base_config()`
4. `style = 'minimal'` resets `signcolumn` and `foldcolumn` back to their minimal defaults

This causes the floating window to lose its left gutter, unlike the split layout which correctly retains these options.

image

## Fix

Move `style = 'minimal'` from `base_config()` into `open_win()` so it only applies during the initial `nvim_open_win()` call, not during subsequent `nvim_win_set_config()` updates.

image

### DIFF
--- a/lua/opencode/ui/float_layout.lua
+++ b/lua/opencode/ui/float_layout.lua
@@ -71,7 +71,6 @@ local function base_config(windows)
     height = height,
     row = resolve_position(float.row, vim.o.lines, height),
     col = resolve_position(float.col, vim.o.columns, width),
-    style = 'minimal',
     border = float.border,
   }
 end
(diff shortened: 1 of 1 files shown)


sudo-tee:
Thanks for the fix.

PR #376, opened 2026-05-18
fix(session): switch session when active child/grandchild's ancestor is deleted

hermandavid:
### Problem

When deleting a parent session from the session picker while switched into a child/grandchild subagent session, two bugs occurred:

1. **No session switch triggered** — the deletion check only compared the active session ID directly against the deleted set, so deleting an ancestor while inside a child session was not detected.

2. **Agent mode stuck on subagent** — when the "no remaining sessions" fallback created a new session, the agent mode was never reset. If the active session was running under a subagent (e.g. `explore`), the new session inherited that mode with no way to switch back.

### Solution

- **Ancestor-aware deletion detection**: a new `_is_session_or_ancestor_deleted` helper walks up the `parentID` chain to detect whether any ancestor of the active session is being deleted.

- **Agent mode reset**: when no remaining sessions exist after deletion, `state.model.clear()` and `agent_model.ensure_current_mode()` are now called before creating the new session — matching the reset already present in `switch_session`.

### Test coverage

- Unit tests for `_is_session_or_ancestor_deleted` (direct, parent, grandparent, sibling, unrelated cases)
- Integration tests verifying session switch on ancestor deletion and agent mode reset when all sessions are deleted

### DIFF
--- a/lua/opencode/ui/session_picker.lua
+++ b/lua/opencode/ui/session_picker.lua
@@ -5,6 +5,28 @@ local util = require('opencode.util')
 local api = require('opencode.api')
 local Promise = require('opencode.promise')
 
+---Check whether any session id in `delete_ids` is the session itself or an ancestor
+---@param session_id string
+---@param delete_ids table<string, boolean>
+---@param all_sessions Session[]
+---@return boolean
+function M._is_session_or_ancestor_deleted(session_id, delete_ids, all_sessions)
+  local session_map = {}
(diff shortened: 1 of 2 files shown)


hermandavid:
Failing CI should be unrelated. There was a regression in nightly neovim that was fixed today.

sudo-tee:
@hermandavid 

Thanks for the fix.

PR #377, opened 2026-05-19
perf(renderer): avoid server round-trip on display toggles and incremental fold updates

CnsMaple:
## Summary

Three performance fixes for the renderer, plus one bug fix for fold ranges not tracking content during streaming.

### 1. Cache-based re-render for display toggles
`toggle_reasoning_output`, `toggle_tool_output`, and `toggle_max_messages` now re-render from `state.messages` in memory instead of fetching the full session from the server. A new `render_from_cache()` path skips `fetch_session()` entirely, making toggles instant (no network round-trip).

### 2. Incremental fold updates during streaming
`update_part_folds(part_id)` replaces `set_all_folds()` in the streaming update path. Instead of rebuilding the entire fold list from all parts on every chunk, it only updates the fold ranges for the single part that changed, then merges with the cached per-part fold list.

### 3. Remove unnecessary `vim.deepcopy`
`_render_full_session_data` was deep-copying the entire messages table on every re-render. All callers pass tables that are either fresh API responses (no other references) or are used synchronously (no concurrent mutation window). The deep copy served no actual protection -- event handlers mutate `state.messages` in-place anyway after storage.

### 4. Fix fold range not updating during append-only streaming
`append_part_now()` wrote new lines but never updated the part's fold range. During streaming, the fold's `to` boundary stayed at the initial value while content grew, causing new content to appear outside the fold. Now calls `update_part_folds()` after appending.

## Verification

Tested on Windows + nvy + Neovim -- the lag that was present when toggling reasoning/tool output visibility and during streaming of large sessions is almost completely gone.

## Notes

- This PR was AI-generated and requires careful review.
- The `render_from_cache` path is safe because it executes synchronously with no yield points -- `_render_full_session_data` is a single contiguous call.
- The incremental fold cache (`ctx.part_folds`, `ctx.global_folds`) is cleared on `ctx:reset()` (full session re-render) and rebuilt from scratch via `set_all_folds()`. No stale cache issues.

Closes #322 (partially)

### DIFF
--- a/lua/opencode/ui/renderer.lua
+++ b/lua/opencode/ui/renderer.lua
@@ -316,11 +316,15 @@ end
 ---Render all messages and parts from session_data into the output buffer
 ---Called after a full session fetch or when revert state changes
 ---@param session_data OpencodeMessage[]
----@param opts? { restore_model_from_messages?: boolean }
+---@param opts? { restore_model_from_messages?: boolean, skip_deepcopy?: boolean }
 function M._render_full_session_data(session_data, opts)
   opts = opts or {}
   M.reset()
-  state.renderer.set_messages(vim.deepcopy(session_data or {}))
+  if opts.skip_deepcopy then
+    state.renderer.set_messages(session_data or {})
+  else
+    state.renderer.set_messages(vim.deepcopy(session_data or {}))
+  end
 
   if not state.active_session or not state.messages then
     return
@@ -399,6 +403,7 @@ function M.render_from_cache(session_data)
   end
   M._render_full_session_data(session_data, {
     restore_model_from_messages = true,
+    skip_deepcopy = true,
   })
   local active_session = state.active_session
   if active_session and active_session.id then
--- a/lua/opencode/ui/renderer.lua
+++ b/lua/opencode/ui/renderer.lua
@@ -316,15 +316,11 @@ end
 ---Render all messages and parts from session_data into the output buffer
 ---Called after a full session fetch or when revert state changes
 ---@param session_data OpencodeMessage[]
----@param opts? { restore_model_from_messages?: boolean, skip_deepcopy?: boolean }
+---@param opts? { restore_model_from_messages?: boolean }
 function M._render_full_session_data(session_data, opts)
   opts = opts or {}
   M.reset()
-  if opts.skip_deepcopy then
-    state.renderer.set_messages(session_data or {})
-  else
-    state.renderer.set_messages(vim.deepcopy(session_data or {}))
-  end
+  state.renderer.set_messages(session_data or {})
 
   if not state.active_session or not state.messages then
     return
@@ -403,7 +399,6 @@ function M.render_from_cache(session_data)
   end
   M._render_full_session_data(session_data, {
     restore_model_from_messages = true,
-    skip_deepcopy = true,
   })
   local active_session = state.active_session
   if active_session and active_session.id then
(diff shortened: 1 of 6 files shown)


CnsMaple:
Additional note from testing:

In sessions with 500k+ context, some lag is still noticeable. The most effective workaround currently is:

`lua
require('opencode').setup({
    ui = {
        output = {
            tools = {
                use_folds = false,
            },
        },
    },
})
`

This PR significantly reduces the lag compared to before, but there is still room for further optimization of the fold/collapse mechanism in very large sessions. The remaining bottleneck appears to be in the fold evaluation itself (Neovim's fold engine evaluating oldexpr per line), which is harder to optimize from the plugin side.

sudo-tee:
Thanks for the PR, I will have a quick review today.

This looks like a really good improvement.

lua/opencode/ui/renderer/buffer.lua
```diff
@@ -635,22 +635,89 @@ end
...
+      end
+    end
+  end
+  if not found then
+    for _, nf in ipairs(new_folds) do
+      table.insert(new_global, nf)
+    end
+  end
+
+  ctx.part_folds[part_id] = new_folds
```
sudo-tee: I think this whole block can be simplified.

By assigning the new folds to the ctx.part_folds first then after this it's only matter of looping to the fold list.


```lua
local new_global = {}

ctx.part_folds[part_id] = new_folds

for pid, part_folds in pairs(ctx.part_folds) do
  for _, f in ipairs(part_folds ) do
    table.insert(new_global, f)
  end
end
```

lua/opencode/ui/renderer/buffer.lua
```diff
@@ -635,22 +635,89 @@ end
...
+    for i = 1, #new_folds do
+      if new_folds[i].from ~= old_folds[i].from or new_folds[i].to ~= old_folds[i].to then
+        same = false
+        break
+      end
+    end
+    if same then
+      return
+    end
+  end
```
sudo-tee: This could become a reusable function `folds_equal`

```lua
local function folds_equal(a, b)
  if #a ~= #b then return false end
  for i = 1, #a do
    if a[i].from ~= b[i].from or a[i].to ~= b[i].to then return false end
  end
  return true
end
```

lua/opencode/ui/renderer.lua
```diff
@@ -390,6 +390,23 @@ function M._render_full_session_data(session_data, opts)
   end
 end
 
+---Re-render from cached session data without a server round-trip.
+---Used for display-only changes (toggle folds, max_messages, etc.)
+---@param session_data OpencodeMessage[]
+function M.render_from_cache(session_data)
```
sudo-tee, line 396: Great idea ,  the round trip was not necessary at all.

sudo-tee (commented):
Great PR, I only added 2 small comments to th `update_part_folds` because it was hard to process with a lot of loops. 

Can you add a test for the update_part_folds logic ? I usually try to incrementally add test coverage throughout the plugin.  

Once this fixed it could be ready to merge.

Thanks a lot

sudo-tee:
@CnsMaple it looks like the tests are failing.

A part from that everything is good

CnsMaple:
All my local tests passed; I might need to trouble you to trigger it again.

sudo-tee:
Thank you for the fixes

PR #379, opened 2026-05-20
fix: display cost during streaming, guard winbar against nil/zero crashes, and async providers for percentage

CnsMaple:
AI-assisted fixes for cost/percentage display, winbar crashes, and mode switch nil check.

## Summary

Three interconnected fixes:

### 1. Cost now appears during streaming (not just after completion)

`lua/opencode/ui/renderer/events.lua`

The `step-finish` SSE part carries token/cost data but was skipped entirely as "bookkeeping". Now extracts token/cost before the early return, so `set_stats` fires during streaming rather than waiting for the final `message.updated`.

### 2. Guard format_token_info against nil/zero crashes

Files: `lua/opencode/ui/topbar.lua`, `lua/opencode/state/renderer.lua`

- `state.cost = 0` is truthy in Lua, causing `if state.cost then` to enter the cost branch. `util.format_cost(0)` returns nil, and `table.insert(parts, nil)` crashes in LuaJIT. Fixed by checking `state.cost > 0`.
- All `table.insert(parts, ...)` calls now guard against nil values.
- `set_stats` batch path also skips zero cost to match `set_cost` behavior.
- Replaced manual space-padding winbar alignment with neovim's `%=` statusline directive for reliable right-alignment regardless of description character width.

### 3. Fix nil check in switch_mode

`lua/opencode/commands/handlers/agent.lua`

`util.index_of` returns `nil` not `-1`. The old check `== -1` never matched, causing `nil % #modes` arithmetic error when `current_mode` isn't in the agents list.

### 4. Percentage now shows (async providers fix)

`lua/opencode/config_file.lua`

`get_model_info` used `:peek()` on the providers promise, returning nil if not yet resolved -> `limit = 0` -> percentage skipped. Added a one-time `finally` callback on the providers promise that triggers `topbar.render()` when providers load, so the percentage appears automatically once the data is available.

### DIFF
--- a/lua/opencode/commands/handlers/agent.lua
+++ b/lua/opencode/commands/handlers/agent.lua
@@ -55,7 +55,7 @@ M.actions.switch_mode = Promise.async(function()
   local modes = config_file.get_opencode_agents():await() --[[@as string[] ]]
   local current_index = util.index_of(modes, state.store.get('current_mode'))
 
-  if current_index == -1 then
+  if current_index == nil then
     current_index = 0
   end
 
--- a/lua/opencode/config_file.lua
+++ b/lua/opencode/config_file.lua
@@ -51,16 +51,27 @@ M.get_workspace_snapshot_path = Promise.async(function()
   return home .. '/.local/share/opencode/snapshot/' .. project.id
 end)
 
+local _providers_render_callback = false
+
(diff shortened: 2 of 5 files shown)


sudo-tee:
Thanks, this seems like a good fix. I didn't have this error myself, but I see why it could happen.

Thanks For the PR

Issue #43, opened 2025-10-03
Planned / In progress features


sudo-tee:
Here is the list of the planned / in progress features for this plugin. 

- [x]  Persistent server running in the background
- [x] Permissions with dialog to approve
- [x]  Commands with arguments
- [x]  Restore mentions context when using the history navigation
- [x] Keymap config overhaul (in progress) by @cameronr 
- [x]  Incremental rendering (in progress) #54 
- [x]  Enqueue messages instead of aborting current request @cameronr #92 
- [x]  mention folders with `@` (planned opencode feature) https://github.com/sudo-tee/opencode.nvim/pull/70

- [x]  Display a context bar to quickly glance and manage the context to be sent (in progress) https://github.com/sudo-tee/opencode.nvim/pull/69

- [x] Make a single Opencode command with sub commands 
- [x]  Add a timeline picker similar to opencode tui
- [x]  Add support for image pasting 

Inline Completion/chat

- [x] Add a way to do 'headless' chat in the current buffer
     - Complete the function around the cursor
     - Do someting on in the visual selection
     - Quick ask on the current buffer
     
     

Long term
- [ ] find a new plug-in name opencode.nvim is already taken 
- [ ]  ACP support ? (https://agentclientprotocol.com/protocol/session-setup) (https://opencode.ai/docs/acp/) ?

There is still a lot to be done to match the opencode CLI.

cameronr:
Just wanted to say that I've really been enjoying the plugin. Great work!

tolgaerdonmez:
For naming the plugin you could focus on the `neovim first, neovim native` aspect of the plugin.

cameronr:
opencode-native.nvim is growing on me :)

cameronr:
As another possible change, what do you think about moving `Opencode*` cmds to be under a single `Opencode` command? That seems to be preferred style to prevent filling up the vim cmd namespace?

sudo-tee:
Yes I agree. It's also a feature I was thinking about. I didn't write it here but this would be a great improvement

cameronr:
@sudo-tee I'm happy to work on that. Assuming that's ok, I'll create an issue to discuss

sudo-tee:
Sure go ahead. It should be more straightforward than incremental rendering 😅. 

For me I really want to finish the context-bar

cameronr:
Great, created #82. Haha, I sure hope it's more straightforward :)

[...]

PR #381, opened 2026-05-20
fix(session): block messaging and hide input for child sessions

hermandavid:
## Problem

Child sessions (sessions with a `parentID`) are read-only views of subagent/task progress. The OpenCode TUI does not allow sending messages to them, users can only observe. The NeoVim plugin had no such restriction: users could navigate to a child session, see the input window, type a message, and submit it.

## Solution

**Block sends at the messaging layer** -- `messaging.send_message` now returns `false` early when `state.active_session.parentID` is set, before reaching `api_client.create_message`. This matches the existing guard pattern for missing active sessions.

**Hide the input window on child session switch** -- `session_runtime.switch_session` now hides the input window and focuses the output window when the target session has a `parentID`. When switching back to a regular session, the input window is restored. Redundant hide/show calls are skipped when the window is already in the correct state.

### DIFF
--- a/lua/opencode/services/messaging.lua
+++ b/lua/opencode/services/messaging.lua
@@ -17,6 +17,10 @@ M.send_message = Promise.async(function(prompt, opts)
     return false
   end
 
+  if state.active_session.parentID then
+    return false
+  end
+
   local mentioned_files = context.get_context().mentioned_files or {}
   local allowed, err_msg = util.check_prompt_allowed(config.prompt_guard, mentioned_files)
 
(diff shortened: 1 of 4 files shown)


sudo-tee:
Good idea.

Thanks for the PR, I like it

jensenojs:
Oh... I sometimes use this feature because occasionally the main agent fails, but a certain subagent's context is well-organized. I don't want to waste it. @sudo-tee , Can I make a switch later?

https://github.com/sudo-tee/opencode.nvim/issues/43#issuecomment-4486450406

sudo-tee:
> Oh... I sometimes use this feature because occasionally the main agent fails, but a certain subagent's context is well-organized. I don't want to waste it. @sudo-tee , Can I make a switch later?
> 
> [#43 (comment)](https://github.com/sudo-tee/opencode.nvim/issues/43#issuecomment-4486450406)
 
@jensenojs 

Yes you are welcome to add a config option to switch the behavior off.

PR #385, opened 2026-05-22
feat(output): add configurable time_format with locale encoding fix

CnsMaple:
## Summary

Add a new `ui.output.time_format` config option to allow users to customize timestamp display format in message headers. Also fixes a locale encoding issue on Chinese Windows where `os.date('%b')` returns GBK-encoded characters (e.g. `5月` → displayed as `5<d4><c2>`).

### `lua/opencode/config.lua`
- Added `time_format = nil` default under `ui.output`

### `lua/opencode/types.lua`
- Added `---@field time_format string|nil` to `OpencodeUIOutputConfig`

### `lua/opencode/util.lua`
- If `time_format` is set, uses it directly via `os.date(time_format, timestamp)`
- When nil (default), wraps locale-sensitive `os.date('%X', ...)` and `os.date('%b', ...)` calls in `os.setlocale('C', 'time')`/restore pairs to prevent non-UTF-8 output

### `tests/unit/util_spec.lua`
- Updated test helpers to use C locale for comparison, matching the new behavior

## Usage

```lua
require('opencode').setup({
  ui = {
    output = {
      time_format = '%m/%d %H:%M',  -- e.g. "05/21 19:13"
    },
  },
})
```

When unset (nil), behavior matches the original format but with the encoding bug fixed.

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -160,6 +160,7 @@ M.defaults = {
     },
     output = {
       filetype = 'opencode_output',
+      time_format = nil,
       compact_assistant_headers = false,
       rendering = {
         markdown_debounce_ms = 250,
--- a/lua/opencode/types.lua
+++ b/lua/opencode/types.lua
@@ -253,6 +253,7 @@
 ---@field event_collapsing boolean
 
 ---@class OpencodeUIOutputConfig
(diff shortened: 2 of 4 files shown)


sudo-tee:
Thanks for the PR

PR #387, opened 2026-05-23
fix(session): prevent input window from reopening in child sessions

hermandavid:
## Problem

In child sessions, the input window is intentionally hidden since users cannot message a child agent. However, pressing `<tab>` (or any keybinding that triggers `focus_input`/`toggle_pane`) while viewing a child session reopens the input window.

## Solution

Add child-session guards at three layers, enforcing the invariant closer to where the input window is actually shown:

- **`input_window._show()`** — safety net at the lowest level. Since `_show()` is the only function that physically recreates the input window, this single guard prevents all current and future callers from accidentally revealing input during a child session.

- **`ui.focus_input()`** — early return so service-layer callers (`agent_model`, `session_runtime`) don't attempt to focus a window that should not exist.

- **`ui.toggle_pane()`** — early return so `<tab>` in the output pane stays on the output instead of trying to show and focus the hidden input.

### DIFF
--- a/lua/opencode/ui/input_window.lua
+++ b/lua/opencode/ui/input_window.lua
@@ -591,6 +591,11 @@ end
 
 ---Show the input window by recreating it
 function M._show()
+  -- Child sessions must never show the input window
+  if state.active_session and state.active_session.parentID then
+    return
+  end
+
   local windows = state.windows
   if not windows or not windows.input_buf or not windows.output_win then
     return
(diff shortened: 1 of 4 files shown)


hermandavid:
Failing CI is a pre-existing issue.

sudo-tee:
Thanks for the PR. For the CI no worries. I know it was broken.

PR #388, opened 2026-05-24
feat(formatter): display loaded skill name in skill tool calls

hermandavid:
## Issue

When the `skill` tool is called during a session, the output shows a generic tool call with no indication of which skill was loaded. The user sees that "skill" was invoked but cannot identify the specific skill (e.g., `commit-message`, `context7-cli`) without inspecting raw data.

## Solution

Add a dedicated `skill` tool formatter that extracts and displays @`input.name`, making the loaded skill immediately visible in the conversation output. Register a distinct icon to visually differentiate skill calls from generic tool calls.


sudo-tee:
This is great. This is the thing I didn't know I was needing 😂


Thanks for another great contribution

PR #390, opened 2026-05-28
feat(session): add child_readonly config and restore/lock agent type for child sessions

jensenojs:
- Add `child_readonly` config (default: `true`) to opt into messaging child sessions
- Infer child session agent type from first assistant message (session.agent is unreliable — reflects last message agent, not creation agent)
- Lock agent type in child sessions, reject user-initiated mode switches

Related: #381, #387

### DIFF
--- a/lua/opencode/services/session_runtime.lua
+++ b/lua/opencode/services/session_runtime.lua
@@ -13,6 +13,26 @@ local agent_model = require('opencode.services.agent_model')
 
 local M = {}
 
+local function focus_after_session_switch(selected_session)
+  if not state.ui.is_visible() then
+    M.open()
+    return
+  end
+
+  if selected_session and selected_session.parentID and config.child_readonly then
+    if not input_window.is_hidden() then
+      input_window._hide()
+    end
+    ui.focus_output()
+    return
+  end
+
+  if input_window.is_hidden() then
+    input_window._show()
+  end
+  ui.focus_input()
+end
+
 ---@param parent_id string?
 M.select_session = Promise.async(function(parent_id)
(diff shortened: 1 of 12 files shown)


lua/opencode/services/session_runtime.lua
```diff
@@ -45,11 +46,63 @@ M.switch_session = Promise.async(function(session_id)
...
+    if vim.tbl_contains(available_agents, selected_session.agent) then
+      state.model.set_mode(selected_session.agent)
+    end
+    if selected_session.model then
+      local model_str = selected_session.model.providerID .. '/' .. selected_session.model.id
+      state.model.set_model(model_str)
+    end
+  else
+    agent_model.ensure_current_mode():await()
+  end
```
sudo-tee: I would split this into multiple little helpers as it's a bit hard to follow, there is a lot of nesting.

sudo-tee (commented):
Looks good.

Only 1 comment about the mode/model resolving.

jensenojs:
I think this should wait for #397. This PR still filters restored `msg.info.mode` through visible agent/subagent lists, but hidden subagents should only be hidden from primary-agent discovery, not from restoring an existing session's agent type. After #397 lands, I can update this PR to keep that restore semantics and split the resolving logic into smaller helpers.

sudo-tee:
> I think this should wait for #397. This PR still filters restored `msg.info.mode` through visible agent/subagent lists, but hidden subagents should only be hidden from primary-agent discovery, not from restoring an existing session's agent type. After #397 lands, I can update this PR to keep that restore semantics and split the resolving logic into smaller helpers.

Make sense to me after having a look at #397

PR #393, opened 2026-05-29
perf(render): fix lazy-render and fold performance for large sessions

### LINKED ISSUE #392: Rendering large sessions is extremely slow (>30s for ~2800 messages)
Switching to a session with ~2800 messages (~9300 parts, ~87000 rendered lines) freezes the UI for 30+ seconds.

Profiled on a session with 2786 messages, 9312 parts, 86946 rendered lines:

| Phase | Time | % |
|---|---|---|
| API fetch | ~1900ms | 5% |
| Format (Lua) | 437ms | 1.2% |
| `nvim_buf_set_lines` | 788ms | 2.2% |
| Extmarks | 69ms | 0.2% |
| **Fold creation** (`cursor` + `normal! zc/zo`) | **35,266ms** | **96.5%** |

The bottleneck is `output_window.set_folds()` — it iterates through fold ranges calling `vim.fn.cursor()` + `vim.cmd('silent! normal! zc')` for each one. With 392 folds in an 87K-line buffer, each `cursor` + `zc` triggers a screen redraw (~90ms per fold). A second pass restoring previously-open folds (`zo`) doubles the cost.

```lua
for _, range in ipairs(folds.ranges) do
  vim.fn.cursor(range.from, 1)
  vim.cmd('silent! normal! zc')
end
```

jensenojs:
## Problem

2786-message sessions take ~36 seconds to render. Two root causes:
1. `set_folds()` uses `cursor() + normal! zc/zo` per fold, each triggering a full-screen redraw (~90ms/fold × 392 folds)
2. `_render_full_session_data()` renders all messages on initial load

Closes #392

### Fold performance (36s → 3s)
- Replace `vim.fn.cursor() + normal! zc/zo` with `:{from},{to}fold` and `:{from},{to}foldopen!` — Ex commands don't move the cursor, zero screen redraws
- Switch to `foldmethod=manual` before creating folds and stay there — prevents foldexpr recalculation for every buffer line (~4.6s on 87K lines)
- Add line count validation to prevent `E16: Invalid range` errors

### Lazy render (3s → ~10ms)
- Only render enough messages to fill the output window on initial load (~1.5× viewport height)
- Load more on scroll-to-top via `WinScrolled` autocmd
- `opts.lazy = true` explicit opt-in — tests and headless environments render all messages
- **Fix `lazy_render_count` being cleared by `M.reset()`**: read before reset, persist back to `ctx` after determining limit
- Debounce `WinScrolled` `load_more` callback (150ms) to prevent rapid re-renders during fast scrolling
- Remove debug `vim.notify` logging from render paths

## Performance

| Phase | Before | After |
|---|---|---|
| Fold creation | ~35,266ms | ~0.8ms |
| Total render | ~36,000ms | ~10ms |
| E2E (incl. async fetch) | ~36,000ms | ~1,875ms |

*Measured on a 2786-msg / 87K-line session*

## Testing

7 new test cases in `tests/unit/renderer_lazy_spec.lua`:
- Renders all messages when `lazy=false`
- Renders limited messages when `ctx.lazy_render_count` is set
- **Preserves `lazy_render_count` increment across render reset** (exposes the core bug)
- `load_more_messages` increments rendered message count
- `load_more_messages` returns false when all/no messages loaded
- No INFO-level debug notifications during rendering

All 110 existing tests pass.

### DIFF
--- a/lua/opencode/ui/output_window.lua
+++ b/lua/opencode/ui/output_window.lua
@@ -593,6 +593,15 @@ end
 function M.setup_keymaps(windows)
   local keymap = require('opencode.keymap')
   keymap.setup_window_keymaps(config.keymap.output_window, windows.output_buf)
+
+  -- When lazy-render is active, gg only reaches the top of rendered content.
+  -- Load all messages first so gg reaches the true start of history.
+  -- See tests/unit/renderer_lazy_spec.lua: "gg loads all messages and makes them searchable"
+  vim.keymap.set('n', 'gg', function()
+    local renderer = require('opencode.ui.renderer')
+    renderer.load_all_messages()
+    return 'gg'
+  end, { buffer = windows.output_buf, expr = true, remap = true })
 end
 
 ---@param windows OpencodeWindowState
@@ -633,13 +642,42 @@ function M.setup_autocmds(windows, group)
     end,
   })
 
-  -- Lazy-render: load more messages on scroll-to-top (debounced)
-  local debounced_load_more = require('opencode.util').debounce(function()
-    local renderer = require('opencode.ui.renderer')
-    if renderer.load_more_messages() then
-      pcall(vim.api.nvim_win_set_cursor, windows.output_win, { 3, 0 })
-    end
-  end, 150)
+    -- Lazy-render: load more messages on scroll-to-top (debounced)
+    local debounced_load_more = require('opencode.util').debounce(function()
+      local renderer = require('opencode.ui.renderer')
+      local render_state = require('opencode.ui.renderer.ctx').render_state
+      -- Save the message at the top of the viewport before loading more
+      local ok, cursor = pcall(vim.api.nvim_win_get_cursor, windows.output_win)
+      local anchor_msg_id = nil
+      local anchor_was_at_line = nil
+      if ok and cursor then
+        local top_line = cursor[1]
+        for _, msg in ipairs(state.messages or {}) do
+          local msg_id = msg.info and msg.info.id or ''
+          if not msg_id:match('^__opencode_') then
+            local rendered = render_state:get_message(msg_id)
+            if rendered and rendered.line_start and rendered.line_start >= top_line then
+              anchor_msg_id = msg_id
+              anchor_was_at_line = rendered.line_start
+              break
+            end
+          end
+        end
+      end
+
+      if renderer.load_more_messages() then
+        -- Restore cursor to the anchor message's new position
+        if anchor_msg_id then
+          local rendered = render_state:get_message(anchor_msg_id)
+          if rendered and rendered.line_start then
+            pcall(vim.api.nvim_win_set_cursor, windows.output_win, { rendered.line_start, 0 })
+            return
+          end
+        end
+        -- Fallback: move to top of buffer
+        pcall(vim.api.nvim_win_set_cursor, windows.output_win, { 1, 0 })
+      end
+    end, 150)
   vim.api.nvim_create_autocmd('WinScrolled', {
     group = group,
     buffer = windows.output_buf,
--- a/lua/opencode/ui/output_window.lua
+++ b/lua/opencode/ui/output_window.lua
@@ -596,12 +596,11 @@ function M.setup_keymaps(windows)
 
   -- When lazy-render is active, gg only reaches the top of rendered content.
   -- Load all messages first so gg reaches the true start of history.
-  -- See tests/unit/renderer_lazy_spec.lua: "gg loads all messages and makes them searchable"
   vim.keymap.set('n', 'gg', function()
     local renderer = require('opencode.ui.renderer')
     renderer.load_all_messages()
-    return 'gg'
-  end, { buffer = windows.output_buf, expr = true, remap = true })
+    vim.api.nvim_win_set_cursor(0, { 1, 0 })
+  end, { buffer = windows.output_buf })
 end
 
 ---@param windows OpencodeWindowState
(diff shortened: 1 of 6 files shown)


sudo-tee:
Thanks for the PR.

I will have a look at soon, I will be out for the weekend. So I will look at it at the start of the next week

jensenojs:
I want to go out and have fun too. Have a nice weekend!

lua/opencode/ui/output_window.lua
```diff
@@ -404,31 +404,28 @@ function M.set_folds(fold_ranges)
...
   vim.api.nvim_win_call(win, function()
     local view = vim.fn.winsaveview()
-    vim.cmd('silent! normal! zx')
-    local prev_starts = {}
-    for _, start_line in ipairs(prev_folds.starts) do
-      prev_starts[start_line] = true
-    end
 
+    -- manual avoids foldexpr recalculation on each fold operation
+    vim.api.nvim_set_option_value('foldmethod', 'manual', { win = 0 })
```
sudo-tee: @jensenojs 

in the setup we set the fold method to `expr`

Image

If we are to manually do it here, I think we could do it manually every time ?

I did some test and removed the expr all together and it seems to work properly.  

In this MR the fold_expr was still called on every line, propable because it was set by default in the setup.

I would move this option in the setup and remove the fold_expr function since with this fix it becomes redundant :)

lua/opencode/ui/renderer.lua
```diff
@@ -12,6 +12,25 @@ local M = {}
 local HIDDEN_MESSAGES_NOTICE_MESSAGE_ID = '__opencode_hidden_messages_notice__'
 local HIDDEN_MESSAGES_NOTICE_PART_ID = '__opencode_hidden_messages_notice_part__'
 
+-- Lazy-render: render only viewport-sized message count on initial load,
+-- load more on scroll-to-top.
+local EST_LINES_PER_MSG = 5
+local VIEWPORT_BUFFER = 1.5
```
sudo-tee: Instead of a comment we could name variables with a `LAZYRENDER_`  prefix.

sudo-tee:
Awesome, I will do some testing on this throughout the day and let you know. Since this is a significant change in the rendering, I want to test it a little bit.

sudo-tee:
@jensenojs 

Thanks for the fix, while we are here can we set foldchars 

```lua
  window_options.set_window_option('fillchars', 'fold:-,foldopen:-,foldclose:+,foldsep:│', windows.output_win)
```

otherwise we see the fold number which looks a bit odd

image

lua/opencode/ui/renderer.lua
```diff
@@ -12,6 +12,23 @@ local M = {}
...
+  local ok, height = pcall(vim.api.nvim_win_get_height, win)
+  if not ok or not height or height <= 0 then
+    return math.huge
+  end
+  return math.ceil(height / LAZYRENDER_EST_LINES_PER_MSG * LAZYRENDER_VIEWPORT_BUFFER)
+end
+
 ---@return integer|nil
 local function get_max_rendered_messages()
   local limit = config.ui and config.ui.output and config.ui.output.max_messages
```
sudo-tee, line 34: I feel that we should combine the max_message behavior with the lazy one, otherwise it might make things confusing. And setting it to nil would essentially disable the lazy rendering ?

sudo-tee:
I did some initial test and it works well.

I do have some comments:
 - When loading more with the line action the cursor is moved back to end of the window 

https://github.com/user-attachments/assets/5d8738a3-61ae-408e-b617-d62207774eda

- The config already expose a max_message and truncate the view, I feel we could combine the behaviors ? if not it might cause confision

8eee6d932e59fb6674a2112d983a9e0bfb3a7b5e7301692466503d5ea5cb6b82

Other than that it looks and feels good.

Thanks for your effort

lua/opencode/ui/renderer.lua
jensenojs: Sorry, I might be a bit busy these two days, and I actually lack some context in understanding the "max_message" feature. I can roughly grasp that the relationship between these two features might be a bit confusing, but I haven't figured it out yet.

For my simple profile, SQLite reading doesn't seem to be the bottleneck. I'll look into why that feature was introduced in the next day or two to make sure I haven't missed any considerations

jensenojs:
lazy-render is no longer a user setting — it's always on. hidden_count only reflects max_messages truncation now.

Behavior changes:
- gg loads all messages before jumping to top (full history searchable)
- load_more preserves cursor position instead of jumping to bottom
- Search scope is rendered content (after gg or scrolling up = everything)
Cursor bug: fixed.

@jemag would be great if you could test this branch — i am not sure if max_messages is still needed for performance on your setup. Rendering should be much faster now.

sudo-tee:
Thanks for the fixes. 

I got an error when using the `gg` keymap though

I also can't reach the top with the arrow keys or the scroll wheel nor `gg`

I don;t have any max_messages set, If I set it I still cant reach the top after loadig more

```
E5108: Lua: ...ts/_nvim/opencode.nvim/lua/opencode/ui/output_window.lua:462: E565: Not allowed to change text or change window
stack traceback:
	[C]: in function 'nvim_buf_set_lines'
	...ts/_nvim/opencode.nvim/lua/opencode/ui/output_window.lua:462: in function 'set_lines'
	...ts/_nvim/opencode.nvim/lua/opencode/ui/output_window.lua:696: in function 'clear'
	...rojects/_nvim/opencode.nvim/lua/opencode/ui/renderer.lua:265: in function 'reset'
	...rojects/_nvim/opencode.nvim/lua/opencode/ui/renderer.lua:342: in function '_render_full_session_data'
	...rojects/_nvim/opencode.nvim/lua/opencode/ui/renderer.lua:434: in function 'render_from_cache'
	...rojects/_nvim/opencode.nvim/lua/opencode/ui/renderer.lua:484: in function 'load_all_messages'
	...ts/_nvim/opencode.nvim/lua/opencode/ui/output_window.lua:602: in function <...ts/_nvim/opencode.nvim/lua/opencode/ui/output_window.lua:600>

```

I did initial debugging and it seems like the `debounced_load_more` is never called because the cursor often cursor at the top is something like 8 or 10 and not 3 or less.. If I put 10 the scrolling works, but I still get the `gg` error.


It seems to be an error because nvim does not allow buffer manipulation during expression evaluation.

changing the keymap to somethingl.ike this seems to work properly 

```lua
vim.keymap.set('n', 'gg', function()
    local renderer = require('opencode.ui.renderer')
    renderer.load_all_messages()
    vim.api.nvim_win_set_cursor(0, { 1, 0 })
end, { buffer = windows.output_buf })
```

jensenojs:
I'll follow up on this tomorrow.

jensenojs:
Scroll-to-top not triggering load_more: the <= 3 check assumes content starts near buffer top, but lazy-render can shift it to line 8+. Gate on whether unrendered messages actually exist instead — when nothing is hidden, the threshold no longer matters.

sudo-tee:
Great, it works really well now.

Thanks for your efforts it's really appreciated

PR #395, opened 2026-06-01
feat(navigation): jump to file at cursor in output window

phanen:
resolve_file_at_cursor parses markdown links, file:// URIs, action
lines, and diff hunks. jump_to_file_at_cursor opens the file in
state.last_code_win_before_opencode (winnr('#') fallback) using a
silent :buffer/:edit helper to avoid re-triggering autocmds.

### DIFF
--- a/lua/opencode/ui/navigation.lua
+++ b/lua/opencode/ui/navigation.lua
@@ -2,6 +2,7 @@ local M = {}
 
 local state = require('opencode.state')
 local renderer = require('opencode.ui.renderer')
+local output_window = require('opencode.ui.output_window')
 
 function M.goto_message_by_id(message_id)
   require('opencode.ui.ui').focus_output()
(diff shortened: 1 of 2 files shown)


lua/opencode/ui/navigation.lua
```diff
@@ -61,4 +62,141 @@ function M.goto_prev_message()
...
+    end)
+    vim.api.nvim_set_current_win(target_win)
+  else
+    open_silent(path)
+    if resolved.line then
+      pcall(vim.api.nvim_win_set_cursor, 0, { resolved.line, 0 })
+    end
+  end
+end
+
```
sudo-tee, line 203: I would split this into multiple methods to make it a bit simpler.

The following code was generated by claude but it gives you an idea on a better seperation of concern.
```lua
local function resolve_path(raw)
  if vim.uv.fs_stat(raw) then return raw end
  local absolute = vim.fn.fnamemodify(raw, ':p')
  if vim.uv.fs_stat(absolute) then return absolute end
  local found = vim.fn.findfile(raw, '.;')
  if found ~= '' then return found end
end

local function open_at(win, path, line)
  local function go()
    open_silent(path)
    if line then pcall(vim.api.nvim_win_set_cursor, 0, { line, 0 }) end
  end
  if win and vim.api.nvim_win_is_valid(win) then
    vim.api.nvim_win_call(win, go)
    vim.api.nvim_set_current_win(win)
  else
    go()
  end
end

local function best_target_win()
  local w = state.last_code_win_before_opencode
  if w and vim.api.nvim_win_is_valid(w) then return w end
  local alt = vim.fn.win_getid(vim.fn.winnr('#'))
  if alt ~= 0 and vim.api.nvim_win_is_valid(alt) then return alt end
end

function M.jump_to_file_at_cursor()
  local resolved = M.resolve_file_at_cursor()
  if not resolved then return end
  local path = resolve_path(resolved.path)
  if not path then return end
  open_at(best_target_win(), path, resolved.line)
end
```

sudo-tee (commented):
I really like this idea. 

Thanks for the PR. 


I left a little comment about the jump function which I find a little complex.

Once the comment addressed it should be good to go

jensenojs:
I also have a PR related to this feature, but it's somewhat similar to the one about automatically jumping to sub-sessions. It doesn't seem to have much to do with the code changes in the current PR. I might be quite busy these days, but I may continue to add related features based on this in the future ?

lua/opencode/ui/navigation.lua
phanen: Thanks for you review. I've split into these small functions. 

When no "best_target_win()" found, "NOP" seems better since opencode_output window is `&winfixbuf`.
```lua
local function open_at(win, path, line)
  if not win or not vim.api.nvim_win_is_valid(win) then
    return
  end
  vim.api.nvim_set_current_win(win)
  open_silent(path)
  if line then
    local buf = vim.api.nvim_win_get_buf(win)
    local line_count = vim.api.nvim_buf_line_count(buf)
    line = math.min(line, line_count)
    pcall(vim.api.nvim_win_set_cursor, win, { line, 0 })
  end
end
```

PR #396, opened 2026-06-01
feat(commands): add clear_selections and clear_files actions

phanen:
Granular context clearing: drop only selections or only mentioned
files instead of wiping the whole context. Exposed as actions and
slash commands.


sudo-tee:
Thanks for the PR. 

I'ts a nice and simple idea :)

PR #397, opened 2026-06-01
fix(session): restore subagent mode from messages

jensenojs:
Hidden subagents are filtered out of the visible primary-agent list so primary agents cannot discover them by default. That visibility rule should not affect restoring an existing session's agent type.

When restoring a session, the tabbar/footer and subsequent message routing should use the agent type recorded on the session messages. Filtering msg.info.mode through the visible primary-agent list caused hidden subagent sessions to keep a stale primary mode in state.current_mode.

### DIFF
--- a/lua/opencode/services/agent_model.lua
+++ b/lua/opencode/services/agent_model.lua
@@ -174,7 +174,15 @@ M.initialize_current_model = Promise.async(function(opts)
           state.model.set_model(model_str)
         end
         if msg.info.mode and state.current_mode ~= msg.info.mode then
-          state.model.set_mode(msg.info.mode)
+          local active_session = state.active_session
+          local should_restore_mode = active_session and active_session.parentID
+          if not should_restore_mode then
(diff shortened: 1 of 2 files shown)


jensenojs:
遊名销慢、prevtous_message 缺说閉、Manifest 缺导入）

--

auditor subagent is hidden

• Missing required fic

sudo-tee:
@jensenojs 

This one is related to #382. This will make primary session inherit hidden subagent if they are the last message in that session.  I don;t know if there is a check we could do here for this ? 

The #382 fixed an issue where mode was set to COMPACTION when reading a compacted session with no way to actually get out of that mode. 

Maybe only do this set_mode when the current session is a child ? What do you think ?

jensenojs:
Good catch. I scoped the unfiltered mode restore to child sessions only, so primary sessions still keep the #382 guard and only restore modes from visible agents.

@sudo-tee

sudo-tee:
Thanks for the fix

PR #401, opened 2026-06-05
fix(session): bypass child-session guard during mode initialization

hermandavid:
## Issue

Switching from a child session to a parent session shows `"Cannot switch agent in child session"` and fails to initialize the agent mode. This happens because `session_runtime.switch_session` calls `state.model.clear()` followed by `agent_model.ensure_current_mode()`. The `clear()` wipes the current mode, and `ensure_current_mode` previously delegated to `switch_to_mode`, which contains a guard that rejects mode changes when the active session has a `parentID`. During session switching, the old child session is still the active session when `ensure_current_mode` runs, so the guard fires incorrectly.

## Solution

Extract the mode+model application logic from `switch_to_mode` into a private `apply_mode` helper. `switch_to_mode` retains the child-session guard for user-initiated mode changes, then delegates to `apply_mode`. `ensure_current_mode` calls `apply_mode` directly, bypassing the guard that is only relevant for interactive mode switching, not system initialization during session transitions.

## Changes

- **Extract `apply_mode` helper** (@`agent_model.lua`) — Private async function that sets the mode on state and resolves the associated model from config (user override, agent config, or global fallback). No session guards; callers are responsible for validation.

- **`switch_to_mode`** (`agent_model.lua`) — Keeps the child-session guard and mode validation, then delegates to `apply_mode` instead of inlining the mode/model logic.

- **`ensure_current_mode`** (`agent_model.lua`) — Selects the default or first available agent, then calls `apply_mode` directly instead of going through `switch_to_mode`.

### DIFF
--- a/lua/opencode/services/agent_model.lua
+++ b/lua/opencode/services/agent_model.lua
@@ -105,6 +105,25 @@ M.cycle_variant = Promise.async(function()
   model_state.set_variant(provider, model, next_variant)
 end)
 
+--- Apply mode and resolve its associated model from config.
+--- No session guards; callers are responsible for validation.
+---@param mode string
+local apply_mode = Promise.async(function(mode)
+  state.model.set_mode(mode)
+  local opencode_config = config_file.get_opencode_config():await() --[[@as OpencodeConfigFile]]
+
+  local agent_config = opencode_config and opencode_config.agent or {}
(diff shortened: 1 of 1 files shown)


sudo-tee:
Thanks for the fix :)

PR #402, opened 2026-06-05
fix(picker): format fzf preview entries to list width

hermandavid:
## Issue

PR #399 fixed the fzf-lua session picker preview gutter, but the width handling for preview mode still used the full picker width for session rows. In practice, that pushed the datetime column off the visible list pane and truncated session rows down to the title and ellipsis.

## Solution

Make preview-aware width handling happen where picker item width is actually determined. The list width is now computed before the picker callback is wrapped, so fzf preview entries use the visible list pane width while the picker window itself still uses the full width.

## Changes

- Move the fzf preview width adjustment into `base_picker.pick()`, where item
  width is passed into the picker callback.
- Remove the ineffective @`opts.width` mutation from `fzf_ui()`, which happened
  too late to affect the rendered session rows.
- Add a regression test that verifies preview-mode entries use the list-pane
  width while fzf still receives the full window width.

### DIFF
--- a/lua/opencode/ui/base_picker.lua
+++ b/lua/opencode/ui/base_picker.lua
@@ -515,17 +515,6 @@ local function fzf_ui(opts)
   local fzf_config = create_fzf_config()
   fzf_config.actions = actions_config
 
-  -- When a preview pane is active, fzf-lua splits the window (default
-  -- right:60% preview, left:40% list). Narrow opts.width so that the
-  -- format function produces entries sized for the list pane, not the
-  -- full window. The format closure reads opts.width on each call.
-  local has_preview = opts.preview and opts.preview ~= 'none' and opts.preview ~= false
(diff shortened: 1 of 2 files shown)


hermandavid:
This is a fix for a bug introduced in #399

phanen:
image
@hermandavid  Is this expected? It's looks a bit weird.. since the right side is empty.

phanen:
Ok... it seems correct, snacks picker also looks like this.

hermandavid:
Well, it's not supposed to like this. The date should be aligned to the right. I tried to test this all the way, but I still can't get it right it seems.

Also, I don't have dates there, only time for some reason, and I don't see any thread previews in your screen.

@phanen

phanen:
I'm using vertical layout, maybe that's not considered in current assumption.

PR #403, opened 2026-06-05
feat(session): cross-project session list and session lock

phanen:
Preserve the active session across DirChanged via a session lock
toggle, and have the session picker list all projects from
/experimental/session when locked (bypassing the api_client's
directory injection).

### DIFF
--- a/lua/opencode/api.lua
+++ b/lua/opencode/api.lua
@@ -53,6 +53,7 @@ local action_groups = {
     rename_session = session.rename_session,
     undo = session.undo,
     fork_session = session.fork_session,
+    toggle_session_lock = session.toggle_session_lock,
   },
 
   diff = {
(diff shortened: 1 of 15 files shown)


phanen:
To elaborate why `/experimental/session` is used:

Regardless of what the client passes for `directory` / `scope`, the server injects `ctx.project.id` inside `listByProject` and only queries sessions for that project (see [`listByProject`](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/session/session.ts#L991-L1044)).

phanen:
~bd7be98743495417aa9dfda97620d907f1c0b2c3~ [9d2eda0](https://github.com/sudo-tee/opencode.nvim/pull/403/commits/9d2eda0025085007f4068faa23008fa4fe320bc4) improve the `format_width` as a follow up of https://github.com/sudo-tee/opencode.nvim/pull/402

cc @hermandavid

hermandavid:
Thank you, I replied to the #402.

phanen:
Ok, I think the picker layout fix is correct for me now. I also test vanilla config, the behavior is not changed.

sudo-tee:
This seems like a reasonnable addition.

Thanks for the PR

PR #406, opened 2026-06-07
fix(renderer): scope shared-server events by session

### LINKED ISSUE #405: Permission/question prompts can leak across sessions with a shared opencode server
I’m using one shared `opencode serve` with both opencode.nvim and the TUI.

Setup:
  - opencode.nvim is attached to session A
  - another TUI is attached to session B
  - both use the same server

If session B asks for a permission/question, that prompt can show up in opencode.nvim for session A.

Expected: opencode.nvim should only show prompts for its active session, including child task sessions that belong to it.

This looks like `/global/event` is broadcast to all clients, and the realtime renderer path is not consistently filtering interactive events by session.

jensenojs:
Route renderer event subscriptions through explicit scope policies so global server events do not leak interactive prompts across sessions.

Centralize request ownership checks in session_scope and reuse them from permission/question restore paths. Keep task child-session tool parts working when child events arrive before the parent task part is indexed.

Fixes #405

### DIFF
--- a/lua/opencode/ui/event_scope.lua
+++ b/lua/opencode/ui/event_scope.lua
@@ -1,0 +1,134 @@
+local state = require('opencode.state')
+local session_scope = require('opencode.ui.session_scope')
+
+local M = {}
+
+local function active_session_id()
+  return state.active_session and state.active_session.id
+end
+
+---@param session_id string|nil
+---@return boolean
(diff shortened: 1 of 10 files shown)


sudo-tee:
This is actually an issue I encountered last week I was planning to look at. 

Thanks for beating me to it :)

PR #407, opened 2026-06-07
feat: generic MCP tool formatter with input preservation

jensenojs:
## Problem

MCP tools (e.g. `sequential-thinking_sequentialthinking`) had no formatter — the default `tool.lua` only renders a title line, not the actual content. Additionally, MCP tool completion events sometimes arrive with an empty `input` table, causing the thought/content to disappear after the tool finishes.

### Generic MCP formatter

Replaces the hardcoded `sequential_thinking.lua` with `mcp.lua` that works for any MCP tool:

- Auto-routes tools with `_` in the name (MCP naming: `{server}_{tool}`)
- Renders known content fields: `thought`, `content`, `text`, `query`, `url`, `input`
- Falls back to JSON encoding for non-standard fields
- Applies `OpencodeHint` dimmed highlight to all MCP output

### Input preservation

Three independent fixes, each covering a different data flow path:

| Layer            | File                | Path                                                             |
| ---------------- | ------------------- | ---------------------------------------------------------------- |
| Event collapsing | `event_manager.lua`   | When throttling emitter merges running+completed into one update |
| Renderer update  | `renderer/events.lua` | When completed event arrives separately with empty input         |
| Test replay      | `helpers.lua`         | When `load_session_from_events` processes events sequentially      |

### `fold_exclude` config

New `ui.output.tools.fold_exclude` option to prevent specific tools from being folded:

```lua
fold_exclude = {
  'bash',                                              -- built-in tool (exact match)
  { server = 'sequential-thinking', tool = 'sequentialthinking' }, -- MCP tool
}
```

Default: sequential-thinking excluded (its output should always be visible).


jensenojs:
图片

sudo-tee:
I really like the fold tool options.

Thanks for another great PR.

PR #410, opened 2026-06-08
fix(picker): restore picker_width=false opt-out in session picker

hermandavid:
### Issue

Setting `picker_width = false` to opt out of width overrides and let the picker backend use its own defaults no longer works for the session picker. Commit `b977136` (#403) introduced `config.ui.picker_width or 100` in @`session_picker.lua`, which silently converts `false` into `100` before `base_picker.pick()` ever sees it. The `false` → `nil` opt-out conversion in `M.pick()` (added in #400) is never reached.

### Solution

Remove the `or 100` fallback so `picker_width` is passed through as-is. `M.pick()` already handles all three cases: a number is used directly, `nil` falls back to `config.ui.picker_width`, and `false` is converted to `nil` to let the backend decide its own window size.


sudo-tee:
Thanks for the fix

PR #412, opened 2026-06-09
(refactor + fix) : Strengthen renderer replay tests and fix output-window regressions

### LINKED ISSUE #408: User message.updated metadata changes should not trigger submit-style force scroll
I ran into a case where the output window jumps back to the bottom while I’m reading older output.

The current renderer behavior appears to treat every `message.updated` whose `info.role == "user"` as submit-style intent:

```lua
if msg.info.role == 'user' then
  state.renderer.set_last_user_message(msg)
  scroll(true, 'user_message_updated')
end
```

That seems too broad. `message.updated` on a user message is not always a new user submission; it can also be a later metadata update for a message that is already rendered.

## Primary issue

For a normal user message, I captured this sequence from the local opencode SQLite event table:

```text
seq 2259  message.updated  role=user, no summary
seq 2268  message.updated  role=user, summary={diffs=[]}
seq 2278  message.updated  role=user, summary={diffs=[]}
```

The persisted message metadata ended up as:

```json
{
  "role": "user",
  "time": { "created": 1780917028109 },
  "agent": "orchestrator",
  "model": {
    "providerID": "openai",
    "modelID": "gpt-5.5",
    "variant": "xhigh"
  },
  "summary": { "diffs": [] }
}
```

`summary.diffs` is really an empty array here, not truncated output. My reading is that it means “this turn has no file diff summary”. That metadata is useful, but it does not mean the user submitted another prompt.

At the moment of the jump, renderer-side debug showed the output was not at the bottom:

```text
cursor={427,0}
visible_bottom=443
line_count=1020
found_before=true
rendered_before=true
inserted_new_message=false
info_equal=true
force=true
reason="role_user_message_updated"
```

So the event was for an existing rendered user message, did not add visible user content, and in that captured case did not change metadata at all. It still forced the output window to the bottom.

## Expected behavior

- New visible user submissions can keep the current submit-follow behavior.
- Existing user message metadata updates should update message state, but should not force-scroll.
- Duplicate `message.updated` snapshots should be scroll-idempotent.
- Being `role=user` should not by itself grant force-scroll authority.

## Related observation

Auto compaction can also create a `role=user` message with a `compaction` part:

```json
{
  "role": "user",
  "summary": { "diffs": [] }
}
```

```json
{
  "type": "compaction",
  "auto": true,
  "overflow": true,
  "tail_start_id": "..."
}
```

Since `opencode.nvim` does not currently render `compaction` parts as visible user text, this can show up as an empty `USER` header. That looks related, but I think the scroll bug is the main thing to fix first: `role=user` is broader than “the user just submitted a visible prompt”.

jensenojs:
## Summary

The replay/window-state work ended up exposing more than the original scroll bug.

With cursor, visible-bottom, and line-count snapshots in place, codex and i fixed six renderer behaviors here:

- metadata-only user message updates no longer steal scroll;
- external TUI user messages no longer force-scroll the nvim output;
- `session.compacted` refreshes the active session;
- compaction markers render as a divider instead of an empty `USER` block;
- compaction summaries render their final text while hiding internal step/reasoning parts;
- lazy loading now triggers from the visible top of the window, not just the cursor line.

Fixes #408

## Tests

- `./run_tests.sh -t replay`
- `./run_tests.sh -t tests/replay/user_message_metadata_scroll_spec.lua`
- `./run_tests.sh -t tests/unit/cursor_tracking_spec.lua`
- `stylua --check lua/opencode/ui/renderer/events.lua tests/helpers.lua tests/manual/regenerate_expected.lua tests/manual/renderer_replay.lua tests/replay/renderer_spec.lua tests/replay/user_message_metadata_scroll_spec.lua`


sudo-tee:
Very nice I will test this throughout the day.

This looks like a massive improvement.

Thanks

sudo-tee:
Been playing with it for most of the day and it's been stable

Thanks for PR

PR #413, opened 2026-06-09
fix(picker): align datetime column consistently in time-based pickers

hermandavid:
## Issue

The time column in session and timeline pickers was sized from each item's own timestamp, so recent sessions showing `10:35` and older ones showing `09 Jun 2024 10:35` had different column widths per row — datetimes were misaligned and older entries overflowed the visible pane with FZF picker.

## Solution

The time column is now sized from the widest timestamp in the current item list, computed once when the picker opens. If all sessions are from today the column is compact; if older sessions are present it expands to fit. On picker reopen the width recomputes from the then-current list.

**Primarily targets fzf-lua.** The column width and separator fixes apply to all backends.

Tested with fzf-lua, Snacks, and Telescope.

### DIFF
--- a/lua/opencode/ui/base_picker.lua
+++ b/lua/opencode/ui/base_picker.lua
@@ -315,9 +315,9 @@ end
 local function fzf_ui(opts)
   local fzf_lua = require('fzf-lua')
 
-  local function finder(fzf_cb, width)
+  local function finder(fzf_cb)
     for idx, item in ipairs(opts.items) do
-      local line_str = opts.format_fn(item, width):to_string()
+      local line_str = opts.format_fn(item):to_string()
 
       -- Prepend index with SOH delimiter for reliable matching
       local indexed_line = tostring(idx) .. '\x01' .. line_str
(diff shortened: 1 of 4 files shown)


sudo-tee:
Thanks for the PR.

With snacks picker the time is still to far right and I don't see it.

7c127e4038ea0e9d78590fd8619bbb39f4562e008d41fdda0318e884bdb716c4

If I scroll right 

bd0426b0cbb5cb648da54ac133266e28153da83b76d8b8e42d2e3e3a582016d6

phanen:
This will move me back to https://github.com/sudo-tee/opencode.nvim/pull/402#issuecomment-4632498563. The date is not a issue for me tbh... fzf have the ability to `toggle-wrap` also text is still searchable via query, but if the text will missed when it's truncated in wrong width. We still have an option to split text into multiple lines (e.g. https://github.com/sudo-tee/opencode.nvim/pull/414)

The issue here is its real width cannot be really determined until fzf buffer is created and we get access from these variable `man fzf | rg 'FZF_.*COL'`, that's my previous `--bind start:+transform` hack do, to defer text format_fn and pass environment variable ctx.

hermandavid:
I'll look into it and try to come up with some robust solution working in general.

PR #414, opened 2026-06-09
fix(picker): avoid each item split into multiple entry

phanen:
read0 can be useful in this case: `echo -en "a\nb" | fzf --read0`

EDIT: this can fix some empty line in `/timeline` picker.


sudo-tee:
Thanks for the PR

PR #420, opened 2026-06-15
feat/jump to user message

phanen:
- **feat(output_window): add jump-to-user-message action with [u/]u keymap**
- **fix(navigation): load all messages before user-message jump under lazy render**


sudo-tee:
This is a nice feature to add. 

Thanks for th PR

PR #421, opened 2026-06-19
fix: close the timer to avoid memory leak

phanen:
I notice my nvim sometime memory leak, not sure if this related but better to `close` the timer if it won't be used again. `close` will `stop` internally.


sudo-tee:
Thanks for the fix

PR #422, opened 2026-06-20
fix(ui): keep permission dialog responsive between sequential prompts

DmitriyKost:
## Summary
Sequential permission prompts could leave the UI stale: after selecting a permission, the permission display could be refreshed while `_processing` was still `true`, causing the next queued permission to skip rendering until another keypress forced a refresh.

- Reorder `on_select`: clear `_processing` **before** `remove_permission` so the next render is not skipped.
- Swap the 50ms `vim.defer_fn` for `vim.schedule`, and add a reentrancy guard at the top of `on_select`.
- Move the post-removal render into `remove_permission` itself, and drop the now-redundant render branch in `on_permission_replied`.

## Test plan
- `./run_tests.sh -t tests/unit/permission_window_spec.lua` - 25/25 pass
- `./run_tests.sh -t tests/unit/permission_integration_spec.lua` - 15/15 pass
- Manually reproduced the stale UI on `6ab7d2f`; gone on this branch

### DIFF
--- a/lua/opencode/ui/permission_window.lua
+++ b/lua/opencode/ui/permission_window.lua
@@ -161,6 +161,8 @@ function M.remove_permission(permission_id)
   else
     M._setup_dialog() -- Setup dialog for next permission
   end
+
+  require('opencode.ui.renderer.events').render_permissions_display()
 end
 
 ---Get currently selected permission (always the first one now)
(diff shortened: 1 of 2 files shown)


sudo-tee:
Thanks for the fix.

Works pretty well

PR #423, opened 2026-06-23
loading animation fixes

phanen:
- **fix(animation): don't drive the spinner from local HTTP activity**
- **fix(animation): hydrate spinner state via sync on (re)attach**

### DIFF
--- a/lua/opencode/api_client.lua
+++ b/lua/opencode/api_client.lua
@@ -195,6 +195,13 @@ function OpencodeApiClient:list_sessions(directory)
   return self:_call('/session', 'GET', nil, { directory = directory })
 end
 
+--- List the current status of all sessions in a workspace.
+--- @param directory string|nil Directory path
+--- @return Promise<{[string]: OpencodeSessionStatusInfo}>
+function OpencodeApiClient:list_session_status(directory)
+  return self:_call('/session/status', 'GET', nil, { directory = directory })
+end
+
 --- List sessions across all projects (experimental global endpoint).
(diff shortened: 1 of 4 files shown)


sudo-tee:
Thanks for the PR, 

This looks like a nice improvement. I will do some testing throughout the day

phanen:
there‘s still edge case spinner won‘t disappear and cannot stably reproduce

sudo-tee:
No stress, I tried it a little bit and for the most part it works pretty well. But I'll wait on merging it.

phanen:
> there‘s still edge case spinner won‘t disappear and cannot stably reproduce

I can be reproduced when spinner should stop during opencode window is hidden, so previous commit forgot to clean up potential stale status in `teardown()` https://github.com/sudo-tee/opencode.nvim/blob/aadc79068a850653d109f3e478b7258b832246b0/lua/opencode/ui/loading_animation.lua#L304-L306

phanen:
Another case may happened when SSE connection is broken unexpectedly. In this case not only animation, the whole render will stop working, maybe we can retry SSE connect on failure and backoff... this may be done in another PR.

sudo-tee:
You are right, we can address this in another PR.

I will do one last test today, and it should be good to go

jensenojs:
@sudo-tee maybe test with this pr hh https://github.com/sudo-tee/opencode.nvim/pull/437

sudo-tee:
Did some tests and it seem pretty stable

Thanks for the PR, it's a great improvement

PR #427, opened 2026-06-26
fix: keep input focused during output auto-scroll

jensenojs:
Fixes output auto-scroll temporarily leaving the opencode input window during internal bottom alignment, which can interrupt IME composition. Keeps the existing bottom alignment behavior and adds a regression test.


sudo-tee:
Thanks for the fix :)

PR #433, opened 2026-06-28
fix(cancel): always call abort_session

phanen:
Problem:
`M.cancel` was gated on `state.active_session and
state.jobs.is_running()`. `state.jobs.is_running()` is true only while
the client has an in-flight HTTP request; once the request returns, the
counter is back to zero. So attaching to a session whose model is
already processing, or toggling off/on with `persist_state`, leaves
cancel with nothing to do.

The counter that drives the 3-strike server restart must keep gating on
`state.jobs.is_running()` — its purpose is to detect the user is trying
to abort an in-flight request and the server is not responding, not to
count cancel attempts on an idle session.

Solution:
Move the `state.jobs.is_running()` gate to wrap only the counter
increment. The cancel itself always runs. `abort_session` is a no-op on
the server for an idle session, so calling it is harmless.

### DIFF
--- a/lua/opencode/services/session_runtime.lua
+++ b/lua/opencode/services/session_runtime.lua
@@ -241,8 +241,10 @@ end
 
 ---@param opts? SendMessageOpts
 M.cancel = Promise.async(function()
-  if state.active_session and state.jobs.is_running() then
-    vim.g.opencode_abort_count = (vim.g.opencode_abort_count or 0) + 1
+  if state.active_session then
+    if state.jobs.is_running() then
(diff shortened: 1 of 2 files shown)


phanen:
> The counter that drives the 3-strike server restart must keep gating on
> `state.jobs.is_running()` — its purpose is to detect the user is trying
> to abort an in-flight request and the server is not responding

llm analysis from git blame. Not sure if this is the designed purpose.

sudo-tee:
This was actually a real issue. The fix is right. No need to gate check. The abort intent is usually something you want to happen immediately.


Thanks for the fix

PR #438, opened 2026-06-30
docs: update Docker spawn example

zerone0x:
## Summary
Fixes #343

Update the README Docker auto-spawn example so it works with the current opencode container image.

## Changes
- Use `https://localhost` for the local Docker server URL.
- Use the official `ghcr.io/anomalyco/opencode:latest` image.
- Pass `serve --port 4096 --hostname '0.0.0.0'` directly to the image instead of repeating the `opencode` command.

## Test Plan
- Verified README Markdown code fences are balanced.
- Checked the updated example contains the requested image/command and no longer contains the old `opencode:latest opencode serve` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -834,7 +834,7 @@ Use `spawn_command` to automatically start your server and `kill_command` to sto
 ```lua
 require('opencode').setup({
   server = {
-    url = 'localhost',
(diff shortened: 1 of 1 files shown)


sudo-tee:
Thanks for the docs update

PR #441, opened 2026-07-02
feat: allow selecting the default variant

### LINKED ISSUE #426: [Feat] select "default" variant
no "default" variant in opencode.nvim
Image

opencode.nvim can detect default variant.
Image

but after switching to other variant, you will lose default variant in opencode.nvim
Image

---

This is not a high-priority requirement. `<M-r>` can be used as a replacement.

zerone0x:
## Summary
- add a `default` entry to the variant picker
- store the default selection as nil so saved model variant overrides are cleared
- keep the UI label as `default` while applying the underlying nil variant value

## Verification
- git diff --check
- ./run_tests.sh -t unit *(blocked: nvim: command not found in this environment)*

Fixes #426

### DIFF
--- a/lua/opencode/services/agent_model.lua
+++ b/lua/opencode/services/agent_model.lua
@@ -39,7 +39,7 @@ function M.configure_variant()
       return
     end
 
-    state.model.set_variant(selection.name)
+    state.model.set_variant(selection.value)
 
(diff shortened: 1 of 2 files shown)


sudo-tee:
Thanks for the PR. 

Looks good to me :)

PR #442, opened 2026-07-03
feat: add_visual_selection_inline should support non-file buffer

phanen:
```
opts.keymap.output_window = {
  ['r'] = { 'add_visual_selection_inline', mode = { 'v' } },
}
```

### DIFF
--- a/lua/opencode/context.lua
+++ b/lua/opencode/context.lua
@@ -181,11 +181,6 @@ end
 function M.build_inline_selection_text(range)
   local buf = vim.api.nvim_get_current_buf()
 
-  if not util.is_buf_a_file(buf) then
-    vim.notify('Cannot add selection: not a file buffer', vim.log.levels.WARN)
-    return nil
-  end
-
   local current_selection = BaseContext.get_current_selection(nil, range)
   if not current_selection then
     vim.notify('No visual selection found', vim.log.levels.WARN)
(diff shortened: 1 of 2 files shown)


sudo-tee:
This is a nice adition, and a workflow I dind't know I needed :)

Thanks again for another great little PR

jensenojs:
Sorry to jump in here. I was browsing recent PRs and was pleasantly surprised by this one. The profiling trick from #450 was also a very useful thing to learn.

This PR made me rethink #318 a bit. Maybe the diff-review workflow I was imagining there was too heavy. What I actually want most of the time is simpler: while looking at any interesting part of Neovim — a code selection, a diff hunk, an opencode output block, a plan, or any temporary buffer — I want to select it, send it into the opencode input, and add a comment around it.

From that angle, `add_visual_selection_inline` feels close to a more general primitive than `add_visual_selection`. It works even when the current buffer is not a file, and the result is visible/editable in the input, which is exactly what I want for “comment on this” workflows.

I’m not sure whether `add_visual_selection` and `add_visual_selection_inline` should be merged. Maybe they should stay as two sinks for the same captured selection:

- persistent structured context: `add_visual_selection`
- editable one-shot prompt text: `add_visual_selection_inline`

One related question: do you use `add_visual_selection` often in your daily workflow? For me, selections added to context are usually useful for only one message. After that I often want them gone, but right now I think I have to clear them manually. That makes the persistent context a little awkward for this kind of workflow.

Maybe a small “one-shot selection / quote selection into input” mechanism would cover most of the use case from #318 without needing a dedicated diffview workflow. @phanen

phanen:
> One related question: do you use `add_visual_selection` often in your daily workflow? For me, selections added to context are usually useful for only one message. After that I often want them gone, but right now I think I have to clear them manually. That makes the persistent context a little awkward for this kind of workflow.

tbh I never use `add_visual_selection` after I know there's add_inline version

sudo-tee:
For my part I do use `add_visual_selection` almost daily.

jensenojs:
> For my part I do use `add_visual_selection` almost daily.

Ah, you probably use it more frequently than I do, so do you think the way it's being constantly cited is reasonable? I'm a bit unsure.

sudo-tee:
They should be cleared once sent, If they are not it's a bug.

PR #443, opened 2026-07-03
fix: add_visual_selection_inline shouldn't trim indent

phanen:
fixup af4e1a8151f4cccc1267d3b6d4caeda858787741

```
  line1
  line2
```
 will be without this commit
```
line1
  line2
```


sudo-tee:
Thanks for the fix

PR #446, opened 2026-07-05
fix: don't trigger ModeChanged

phanen:
This seems to break my input method switching plugin


sudo-tee:
There seem to be another attempt at this #452 

@phanen  the other MR fix your issue ?

phanen:
I forgot `:normal` can trigger ModeChanged. so that seems more correct fix.

PR #447, opened 2026-07-05
fix(renderer): sanitize embedded newlines in Output.add_line

phanen:
```
Problem: when the LLM emits question options (or any other
server-supplied text) whose label or description contains a literal
newline, format_options inserts the raw string as a single line in
the formatted output. nvim_buf_set_lines then rejects the line
with 'replacement string item contains newlines' and the whole
flush errors out, dropping the part from the rendered buffer and
leaving an apparently-broken dialog (only the title text visible).

Solution: replace any embedded \r\n, \n, or \r with a single
space at the Output layer. This is the natural defensive boundary:
every line that flows to nvim_buf_set_lines goes through add_line
or add_lines, so sanitizing there covers every renderer call site
(question options, permission titles, MCP tool output, etc.) at
once without scattering special-case handling.
```

To reproduce, maybe just ask opencode to:
```
Use question tool call with newline in string param to test if it work well
```

### DIFF
--- a/lua/opencode/ui/output.lua
+++ b/lua/opencode/ui/output.lua
@@ -35,11 +35,22 @@ function Output.new()
 end
 
 ---Add a new line
----@param line string
----@return number index The index of the added line
+---@param line string|nil
+---@return number index of the first line added (0 if line was nil)
 function Output:add_line(line)
-  table.insert(self.lines, line)
-  return #self.lines
+  if line == nil then
+    table.insert(self.lines, '')
+    return #self.lines
+  end
+  local segments = vim.split(line, '\r?\n')
+  local first_idx = 0
+  for _, segment in ipairs(segments) do
+    table.insert(self.lines, segment)
+    if first_idx == 0 then
+      first_idx = #self.lines
+    end
+  end
+  return first_idx
 end
 
 ---Get line by index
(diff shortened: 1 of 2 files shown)


phanen:
Not sure if there any best way to fix. I originally notice question dialog won't show up if some error happened before that.

Maybe a schema error event in opencode side will produce another bug, I haven't confirm.

lua/opencode/ui/output.lua
```diff
@@ -30,11 +30,28 @@ function Output.new()
...
+---server-supplied strings (question labels/options, permission
+---titles, etc.) cannot crash the renderer when an LLM produces a
+---multiline value.
+---@param line string|nil
+---@return string
+function Output._sanitize_line(line)
+  if line == nil then return '' end
+  local s = tostring(line)
+  s = s:gsub('\r\n', ' ')
+  s = s:gsub('[\n\r]', ' ')
```
phanen: No... this definitely isn't a good solution. I'll look into it later

phanen:
The prompt to test: `Insert newline characters into the string parameters of the question tool call and run a test, aggressively testing as many cases as possible.`

sudo-tee:
The fix make sense to me.

Thanks for the PR

PR #449, opened 2026-07-07
fix(autocmd): skip non-normal buftype for performance

phanen:
I notice `:terminal` becuase obviously slow after eebda20e320100c481ab33b87571481dbab99371

Not sure why `M.invalidate_reference_targets_for_file_change()` is slow, but I think no problem to skip special buffer here.


phanen:
image
before/after `:terminal`

phanen:
oops created from main branch,

PR #450, opened 2026-07-07
fix(autocmd): skip non-normal buftype for performance

phanen:
I notice :terminal becuase obviously slow after eebda20

Not sure why M.invalidate_reference_targets_for_file_change() is slow, but I think no problem to skip special buffer here.

image
before/after `:terminal`

### DIFF
--- a/lua/opencode/ui/autocmds.lua
+++ b/lua/opencode/ui/autocmds.lua
@@ -43,7 +43,7 @@ function M.setup_autocmds(windows)
     group = group,
     pattern = '*',
     callback = function(args)
-      if args.file == '' then
+      if args.file == '' or vim.bo[args.buf].buftype ~= '' then
         return
       end
       require('opencode.ui.renderer.events').invalidate_reference_targets_for_file_change()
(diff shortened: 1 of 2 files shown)


sudo-tee:
@phanen  thanks for the fix.

How did you get that nice flamechart from your nvim config ?

phanen:
```lua
M.profile = function(out)
  out = out or '/tmp/tmp/profile.log'
  fn.mkdir(fs.dirname(out), 'p')
  if not _G.jit_profile then
    -- G option for https://github.com/jonhoo/inferno
    require('jit.p').start('10,i1,s,m0,G', out)
    print('profile started')
  else
    require('jit.p').stop()
    fn.system(('inferno-flamegraph %s > %s'):format(out, fs.joinpath(fs.dirname(out), 'profile.svg')))
    vim.system({ vim.env.BROWSER, '/tmp/tmp/profile.svg' })
    print('profile stoped')
  end
  _G.jit_profile = not _G.jit_profile
end
```

PR #452, opened 2026-07-07
fix(ui): restore output views without normal commands

jensenojs:
## Summary

Remove Normal-mode view commands from automatic output view maintenance.

The previous fix in #427 tried to protect the IME-sensitive output auto-scroll path by suppressing focus-related window/buffer events around `normal! zb`. With the view restore moved to `winsaveview()` / `winrestview()`, that wrapper no longer has a job.

This patch removes the source instead:
- output auto-scroll no longer uses `normal! zb`;
- lazy-render top-anchor restore no longer uses `normal! zt`;
- the #427 `eventignore` wrapper is removed;
- the behavior is covered directly: input focus stays put, and auto-scroll does not emit `WinEnter`, `WinLeave`, `BufEnter`, or `ModeChanged`.

Other `normal!` call sites are left unchanged because they are user navigation / picker / selection paths, not automatic output view maintenance.

## Tests
- `./run_tests.sh`

### DIFF
--- a/lua/opencode/ui/output_window.lua
+++ b/lua/opencode/ui/output_window.lua
@@ -209,7 +209,7 @@ function M.restore_view_topline(win, topline)
     view.topline = math.max(1, topline)
     view.skipcol = 0
     vim.fn.winrestview(view)
-    vim.fn.line('w$')
+    -- Apply the restored viewport before later line('w$') reads inspect it.
     vim.cmd('redraw')
   end)
 end
--- a/lua/opencode/ui/output_window.lua
+++ b/lua/opencode/ui/output_window.lua
@@ -201,6 +201,19 @@ function M.get_visible_top_line(win)
   return (ok and line and line > 0) and line or nil
 end
 
+---@param win integer
+---@param topline integer
+function M.restore_view_topline(win, topline)
+  vim.api.nvim_win_call(win, function()
+    local view = vim.fn.winsaveview()
+    view.topline = math.max(1, topline)
+    view.skipcol = 0
+    vim.fn.winrestview(view)
+    vim.fn.line('w$')
+    vim.cmd('redraw')
+  end)
+end
+
 ---@param win? integer
 function M.reset_scroll_tracking(win)
   if win then
@@ -763,9 +776,7 @@ function M.setup_autocmds(windows, group)
         if rendered and rendered.line_start then
           local restored_top = math.max(1, rendered.line_start + anchor_offset)
           pcall(vim.api.nvim_win_set_cursor, windows.output_win, { restored_top, 0 })
-          pcall(vim.api.nvim_win_call, windows.output_win, function()
-            vim.cmd('normal! zt')
-          end)
+          pcall(M.restore_view_topline, windows.output_win, restored_top)
           return
         end
       end
--- a/tests/unit/cursor_tracking_spec.lua
+++ b/tests/unit/cursor_tracking_spec.lua
@@ -566,13 +566,7 @@ describe('renderer.scroll_to_bottom', function()
       end,
     })
 
-    local cmd_stub = require('luassert.stub')(vim, 'cmd').invokes(function(cmd)
-      if cmd == 'normal! zb' then
-        vim.api.nvim_exec_autocmds('ModeChanged', { pattern = 'i:n', modeline = false })
-        return
-      end
-      return vim.api.nvim_cmd(vim.api.nvim_parse_cmd(cmd, {}), {})
-    end)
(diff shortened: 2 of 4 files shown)


phanen:
Here's another way to avoid emit `ModeChanged`
```lua
local function scroll_view(target)
  local view = vim.fn.winsaveview()
  view.topline = view.lnum - target + 1
  vim.fn.winrestview(view)
end

vim.keymap.set('n', '<leader>zb', function()
  scroll_view(vim.api.nvim_win_get_height(0))
end)
vim.keymap.set('n', '<leader>zz', function()
  scroll_view(math.ceil(vim.api.nvim_win_get_height(0) / 2))
end)

vim.keymap.set('n', '<leader>zt', function()
  scroll_view(1)
end)
```

phanen:
All of this if-else branch should be included if we want to suppressed ModeChanged.
https://github.com/sudo-tee/opencode.nvim/blob/af18aec264358673260db3ef1487fbae8705dfc8/lua/opencode/ui/renderer/scroll.lua#L128-L138

Also there're many other call site not sure if all of them related. 
image

jensenojs:
Thanks, your suggestion is the right direction. I revised the PR to use `winsaveview()` / `winrestview()` for output view restore instead of suppressing `ModeChanged`.

This removes the `normal! zb` / `normal! zt` calls from automatic output maintenance, and also removes the old eventignore wrapper.

Could you give the updated branch a try? @phanen

phanen:
yeah, I'm using this branch now.

lua/opencode/ui/output_window.lua
```diff
@@ -201,6 +201,19 @@ function M.get_visible_top_line(win)
...
+---@param win integer
+---@param topline integer
+function M.restore_view_topline(win, topline)
+  vim.api.nvim_win_call(win, function()
+    local view = vim.fn.winsaveview()
+    view.topline = math.max(1, topline)
+    view.skipcol = 0
+    vim.fn.winrestview(view)
+    vim.fn.line('w$')
+    vim.cmd('redraw')
```
phanen: Are these two lines useful? Looks like nothing has been done.. Or is it to fix certain rendering issues?

jensenojs: Good catch. `line('w$')` is unnecessary, but `redraw` is needed here: without it, replay reads a stale visible_bottom after winrestview(). I removed the extra `line('w$')` and left a comment for `redraw`.

tests/unit/cursor_tracking_spec.lua
```diff
@@ -545,26 +533,64 @@ describe('renderer.scroll_to_bottom', function()
...
       end,
     })
 
+    local cmd_stub = require('luassert.stub')(vim, 'cmd').invokes(function(cmd)
+      if cmd == 'normal! zb' then
+        vim.api.nvim_exec_autocmds('ModeChanged', { pattern = 'i:n', modeline = false })
+        return
+      end
+      return vim.api.nvim_cmd(vim.api.nvim_parse_cmd(cmd, {}), {})
+    end)
```
phanen: Maybe there's no need to handle 'normal!' zb' specially since it will already emit autocmd if it's used.

```diff
-    local cmd_stub = require('luassert.stub')(vim, 'cmd').invokes(function(cmd)
-      if cmd == 'normal! zb' then
-        vim.api.nvim_exec_autocmds('ModeChanged', { pattern = 'i:n', modeline = false })
-        return
-      end
-      return vim.api.nvim_cmd(vim.api.nvim_parse_cmd(cmd, {}), {})
-    end)
+    local cmd_stub = require('luassert.stub')(vim, 'cmd')
```

phanen:
Other thing lgtm, I've been on this branch a day and didn't reproduce input switch issue.

sudo-tee:
Thanks for the PR.

I've tested and it seems pretty stable for me to.

lua/opencode/ui/output_window.lua
```diff
@@ -201,6 +201,19 @@ function M.get_visible_top_line(win)
...
+---@param win integer
+---@param topline integer
+function M.restore_view_topline(win, topline)
+  vim.api.nvim_win_call(win, function()
+    local view = vim.fn.winsaveview()
+    view.topline = math.max(1, topline)
+    view.skipcol = 0
+    vim.fn.winrestview(view)
+    -- Apply the restored viewport before later line('w$') reads inspect it.
+    vim.cmd('redraw')
```
phanen, line 213: This seems causing UAF when using tabpage. I create a pr upstream https://github.com/neovim/neovim/pull/40731

PR #453, opened 2026-07-08
fix/fill input

phanen:
- **feat(session): refill prompt input after /undo**
- **refactor(history): derive prompt history from active session messages**

Not sure why all prompt history is stored in ~/.local/share/nvim/opencode/history.txt.

Ideally different session should have separate prompt history.

Maybe history can be deprecated in favor of timeline picker?

### DIFF
--- a/lua/opencode/commands/handlers/session.lua
+++ b/lua/opencode/commands/handlers/session.lua
@@ -70,10 +70,18 @@ end
 
 ---@param request_promise Promise<any>
 ---@param error_prefix string
-local function run_api_action_with_checktime(request_promise, error_prefix)
-  request_promise:and_then(schedule_checktime):catch(function(err)
-    notify_error(error_prefix, err)
-  end)
+---@param on_success? fun(...)
+local function run_api_action_with_checktime(request_promise, error_prefix, on_success)
+  request_promise
+    :and_then(function(...)
+      schedule_checktime()
+      if on_success then
+        on_success(...)
+      end
+    end)
+    :catch(function(err)
+      notify_error(error_prefix, err)
+    end)
 end
 
 function M.actions.open_input_new_session()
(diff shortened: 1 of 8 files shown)


sudo-tee:
@phanen 

The history file was a remnant from the original repo this one was forked from (goose.nvim)

But just scope it from the session would limit what you can do with it. 

Sometimes I start new sessions but want to send a prompt I already sent in another.

But they should be scoped per project at least.

phanen:
> But they should be scoped per project at least.

This pr currently make history picker per-session since it's just wrapper on timeline picker. Maybe we can have a action for history to toggle session/project/...

Also maybe each session should have persistent draft message if tui support this (i don't often use tui so i'm not sure)

sudo-tee:
@phanen 

Can we split the 2 things that this PR addresses ?

I really like the /undo fix.

But for the history I would need to play with it and think about what is the best solution. 

Because opencode tui does the same history navigation as we do with arrows in the input. and it does not scope it per session either. The history picker in this plugin was just a view on the message history the user sent over time.

phanen:
Found another bug... `/undo` say  "No user message to undo" but it actually have.

msg is often truncated from the end https://github.com/sudo-tee/opencode.nvim/blob/eebda20e320100c481ab33b87571481dbab99371/lua/opencode/ui/renderer.lua#L368-L370
https://github.com/sudo-tee/opencode.nvim/blob/eebda20e320100c481ab33b87571481dbab99371/lua/opencode/ui/renderer.lua#L87-L101


`set_last_user_message` is only called in `on_message_updated` (only on truncated messages)
https://github.com/sudo-tee/opencode.nvim/blob/eebda20e320100c481ab33b87571481dbab99371/lua/opencode/ui/renderer.lua#L375-L377
and only when role is user
https://github.com/sudo-tee/opencode.nvim/blob/eebda20e320100c481ab33b87571481dbab99371/lua/opencode/ui/renderer/events.lua#L291-L292

but a canceled user messages sequence can be:
```
user -> assistant (tool call 1) -> assistant (tool call 2) -> ...
```

so when there's many tool call, truncated msg won't contain user msg -> last_user_message will be never set.

so sometimes `/undo` cannot found last_user_message (usually reproduce on startup or when use `max_messages`, and haven't `load_all_message`/`load_more_message`).

phanen:
`on_message_updated` seems correct to call `set_last_user_message` when it's call from opencode event. But it have have side effect on render job... so `_render_full_session_data` lazily use it... but `set_last_user_message` is check/updated on every message. 

Not sure if it's proper to add another set_last_user_message here... or make anther version of `on_message_updated` without side effect.. i'll think about it later.

EDIT: llm suggest a another option to remove `set_last_user_message` totally and get last userr msg from `store.messages` when require..

phanen:
> EDIT: llm suggest a another option to remove `set_last_user_message` totally and get last userr msg from `store.messages` when require..

I finally choose this direction since it feels proper and easy to go...

sudo-tee:
Since we can restore running sessions, and the output is sliced, it might be a good isea to reconstruct from the messages instead set_last_user_messsage.  it's worth a try at least

sudo-tee:
Looks good to me.

Thanks for the PR

Issue #451, opened 2026-07-07
[chore/feature] : Tell OpenCode what to do differently


jensenojs:
Currently, in the TUI, if we reject a patch, it prompts us to enter a reason; however, in this plugin, it simply interrupts the process. It would be great if we could have a similar experience here

<img width="2236" height="1104" alt="Image" src="https://github.com/user-attachments/assets/ea729705-5c13-497d-b7ee-34f7c6501d56" />

---

<img width="766" height="635" alt="Image" src="https://github.com/user-attachments/assets/2938a897-a5df-4dc0-9d9b-3bb6cdb1bb71" />

sudo-tee:
Good idea. 

I would like to do one QoL refactrore before this. Where you can have an inline input when selecting the "Other" option.

This would fit well in this workflow.

I will have a jab at it and sede if it's possible

disrupted:
sorry for the low effort comment, but I do actually prefer our current behavior. I was recently quite bummed when I used the TUI and didn't find an obvious way to halt the execution completely (circumventing this dialog). is there even a functional difference to cancelling and submitting a new prompt?

jensenojs:
Aha, glad to see you again. However, I'm still curious why this interaction method isn't always present in TUI. It doesn't seem to trigger very stably. But we should always be able to make it optional, right?

sudo-tee:
> sorry for the low effort comment, but I do actually prefer our current behavior. I was recently quite bummed when I used the TUI and didn't find an obvious way to halt the execution completely (circumventing this dialog). is there even a functional difference to cancelling and submitting a new prompt?

You can have a look, it is still "floating window" and work just like snacks input or any other plugin that overrides vim.ui.input, it's just inline.

there is an option to opt-out of it.

https://github.com/sudo-tee/opencode.nvim/pull/455

<img width="827" height="106" alt="Image" src="https://github.com/user-attachments/assets/13ca4575-33bc-4270-8c3d-b7f7d3be38f0" />

sudo-tee:
@disrupted  
I think I misunderstood your comment.  

But yes if we add this interaction we can add an option. We could also support sending an empty reason or even add a `Deny and stop` option

disrupted:
[...]

PR #455, opened 2026-07-08
feat(ui): inline floating input for Other option

sudo-tee:
This PR introduce an inline input for answering the Other question inline.

image

@jensenojs 

If you have a minute. I'd like your `input` (pun intended 😅)  on this. 

Does it work for you, do you think it's a proper implementation that we can use for the Permission `Deny`

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,7 @@ require('opencode').setup({
     },
     questions = {
       use_vim_ui_select = false, -- If true, render questions/prompts with vim.ui.select instead of showing them inline in the output buffer.
+      inline_other_input = true, -- If true, show an inline floating input for "Other" instead of vim.ui.input.
     },
     output = {
       filetype = 'opencode_output', -- Filetype assigned to the output buffer (default: 'opencode_output')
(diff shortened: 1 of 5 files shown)


sudo-tee:
This can help with #451

jensenojs:
It looks great! I'll give it a try in about an hour or two.

jensenojs:
图片

I really like it! But I'm not sure if this is related to this PR, but is our original interaction logic for multi-selection similar to the TUI? It seems that <Enter> directly confirms the selection?

---

and WinLeave drops typed input on any focus loss, my agent said that root cause is using WinLeave as cancel trigger when WinClosed/Esc already covers cleanup.  @sudo-tee

sudo-tee:
Not sure I follow here.

WinLeave is only on the inline input element. 

I don;t think we support multi-selection yet in our imnplementation

jensenojs:
iShot_2026-07-09_09 52 23

Sorry, I should have recorded a GIF, but when the input field is accidentally dismissed and then reopened, the original input is discarded. Perhaps this could be optimized? Multi-selection can be discussed later

sudo-tee:
Oh I see. I will have a look.

Thanks

sudo-tee:
I pushed a fix for the accidental dismiss.

PR #456, opened 2026-07-09
fix(navigation): preserve jumplist before output buffer jumps

phanen:
```
Problem: programmatic cursor moves via nvim_win_set_cursor in the
output window (gg, ]], [[, ]u, [u, <CR>, ...) do not update the
previous-position mark, so users cannot return to the prior output
location with `` or <C-o> after a jump.

Solution: introduce mark_jump_position() which records the current
cursor as the ' mark inside the source window via
`noau normal! m'`, and call it before every nvim_win_set_cursor in
navigation.lua and before the gg keymap in output_window.lua.
```

### DIFF
--- a/lua/opencode/ui/navigation.lua
+++ b/lua/opencode/ui/navigation.lua
@@ -4,6 +4,13 @@ local state = require('opencode.state')
 local config = require('opencode.config')
 local renderer = require('opencode.ui.renderer')
 
+---@param win integer
+local function mark_jump_position(win)
+  pcall(vim.api.nvim_win_call, win, function()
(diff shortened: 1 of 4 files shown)


sudo-tee:
Thanks for the PR.

This is another great QOL improvement

phanen:
test seems flaky on master https://github.com/sudo-tee/opencode.nvim/actions/runs/29008327834/job/86085345936

sudo-tee:
Yes, it's always the same MacOs test that times out.. I will investigate

sudo-tee:
I pushed a fix, to use a fake timer instead of relying on vim.wait

PR #457, opened 2026-07-09
fix(quick_chat): make spinner relative to buffer

phanen:
This will pin the spinner where it was used, but not follow the cursor.


sudo-tee:
Thanks for the fix :)

PR #460, opened 2026-07-10
feat/client/undo

phanen:
Attempt to make it possible `/undo` queued message without abort running session.
(which is not allowed in opencode tui)

This pr is just to show a rough idea and this is possible if we do in client part.


phanen:
The working principle is that the client maintains a separate queue of unsent messages, and only sends messages when the server can process them immediately. This way, the client '/undo' can avoid interrupting server conversations.

I'm just testing the idea, and personally, I don't think it's worth merging, because it's not a very necessary use case and it harms persistence and multi-client synchronization.

I'm just very sorry about OpenCode lacks such a basic feature

PR #461, opened 2026-07-11
fix(renderer): skip hidden-messages notice in message navigation

phanen:
Problem: when max_messages truncates older messages, the synthetic
hidden-messages notice (appended to state.messages via
on_message_updated) is matched by get_prev_rendered_message, so [[
yanks the cursor to the buffer top.

Solution: skip entries where is_renderer_synthetic_message is true in
both get_prev_rendered_message and get_next_rendered_message.

### DIFF
--- a/lua/opencode/ui/renderer.lua
+++ b/lua/opencode/ui/renderer.lua
@@ -619,10 +619,12 @@ function M.get_next_rendered_message(current_line)
   local next_message = nil
 
   for _, message in ipairs(state.messages or {}) do
-    local rendered = message.info and message.info.id and ctx.render_state:get_message(message.info.id) or nil
-    if rendered and rendered.line_start and rendered.line_start + 1 > current_line then
-      next_message = rendered
-      break
+    if not is_renderer_synthetic_message(message) then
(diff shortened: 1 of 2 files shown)


lua/opencode/ui/renderer.lua
```diff
@@ -619,10 +619,12 @@ function M.get_next_rendered_message(current_line)
   local next_message = nil
 
   for _, message in ipairs(state.messages or {}) do
-    local rendered = message.info and message.info.id and ctx.render_state:get_message(message.info.id) or nil
-    if rendered and rendered.line_start and rendered.line_start + 1 > current_line then
-      next_message = rendered
-      break
+    if not is_renderer_synthetic_message(message) then
```
jensenojs, line 622: It looks like this only excludes revert and hidden notices, permission/question displays are also synthetic entries in `state.messages` and may still be picked up by navigation ?

sudo-tee:
seems fair to me. 

Thanks

lua/opencode/ui/renderer.lua
sudo-tee: Good observation. I would have to test such scenario . This might be ok that we can navigate to permissions and questions.

PR #462, opened 2026-07-11
feat(models-picker): use vim.ui.select as models picker

### LINKED ISSUE #459: use vim.ui.select for models picker
currently, i use custom implementation of vim.ui.select.
```lua
vim.ui.select = function()
-- custom implementation that open a simple floating window
end
```
the picker works for /sessions and /agent slash commands, but for /models, it does nothing, i.e. no picker shows up.

pirey:
Attempt to fix #459 

This PR contain change in the base_picker to use vim.ui.select.

The vim.ui.select call is generic and the implementation detail of the custom vim.ui.select is left to the users.

### DIFF
--- a/lua/opencode/commands/handlers/workflow.lua
+++ b/lua/opencode/commands/handlers/workflow.lua
@@ -223,6 +223,9 @@ function M.actions.mention_file()
   local context = require('opencode.context')
   require('opencode.ui.mention').mention(function(mention_cb)
     picker.pick(function(file)
+      if not file then
+        return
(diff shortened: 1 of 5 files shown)


pirey:
https://github.com/user-attachments/assets/a226c6ed-2a0e-4bc4-8ccd-b16a8934c0b0

sudo-tee:
Thanks for the PR. 

This looks like a simple fix for adding support of vim.ui.select. 

Thanks for your contribution

PR #463, opened 2026-07-11
refactor(renderer): remove dead pinned-overlay scroll short-circuit

phanen:
This never work since `a and nil or b` is always `b`

This seems a defensive programming introduced in a52636b47621d24b287ab6302088aa3d114fdcd0

"only once scroll() on dialog first appear" have been implemented here:
https://github.com/sudo-tee/opencode.nvim/blob/3f1baf43ae5011216d147878f7fdc939b38ce226/lua/opencode/ui/renderer/events.lua#L156
https://github.com/sudo-tee/opencode.nvim/blob/3f1baf43ae5011216d147878f7fdc939b38ce226/lua/opencode/ui/renderer/events.lua#L176-L178


phanen:
will push more minor fix here

sudo-tee:
Nice cleanup. 

Thanks for the effort. I tested it and it works great :)

PR #465, opened 2026-07-12
feat: Support auth on remotes + Set password on local

mikew:
Noticed a comment about [eventually looking in to support for username / passwords on remotes](https://github.com/sudo-tee/opencode.nvim/pull/295#issuecomment-3992146786), and since have my opencode behind a username / password, decided to add support for it.

All tests pass and it works locally.

https://github.com/user-attachments/assets/f2929efc-bcf6-4d00-8866-789dd240777f

### DIFF
--- a/lua/opencode/auth.lua
+++ b/lua/opencode/auth.lua
@@ -2,6 +2,7 @@ local config = require('opencode.config')
 
 local M = {}
 
+---@type {password: string | nil, username: string} | nil
 local cache = nil
 
 --- Resolve a credential value that may be a string or a function returning a string.
@@ -26,20 +27,15 @@ end
 ---@return string|nil password
 ---@return string username
 local function ensure_resolved()
-  if cache then
-    return cache.password, cache.username
+  if cache == nil then
+    local password = resolve_credential(config.server.password) or vim.env.OPENCODE_SERVER_PASSWORD
+    local username = resolve_credential(config.server.username) or vim.env.OPENCODE_SERVER_USERNAME or 'opencode'
+
+    cache = {
+      password = password,
+      username = username,
+    }
   end
-
-  local password = resolve_credential(config.server.password)
-    or vim.env.OPENCODE_SERVER_PASSWORD
-  local username = resolve_credential(config.server.username)
-    or vim.env.OPENCODE_SERVER_USERNAME
-    or 'opencode'
-
-  cache = {
-    password = password,
-    username = username,
-  }
 
   return cache.password, cache.username
 end
--- a/lua/opencode/auth.lua
+++ b/lua/opencode/auth.lua
@@ -1,0 +1,63 @@
+local config = require('opencode.config')
+
+local M = {}
+
+local cache = nil
+
+--- Resolve a credential value that may be a string or a function returning a string.
+--- Returns nil for nil, empty string, or function errors.
+---@param val string | (fun(): string | nil) | nil
+---@return string | nil
+local function resolve_credential(val)
+  if type(val) == 'function' then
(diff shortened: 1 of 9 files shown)


lua/opencode/types.lua
```diff
@@ -206,6 +206,8 @@
 ---@field auto_kill boolean -- Kill spawned servers when nvim exits (default: true)
 ---@field path_map (string | fun(host_path: string): string) | nil -- Map host paths to server paths
 ---@field reverse_path_map (fun(server_path: string): string) | nil -- Map server paths back to host paths
+---@field username? string | fun(): string | nil -- Username for Basic auth. Falls back to OPENCODE_SERVER_USERNAME env var, then "opencode"
+---@field password? string | fun(): string | nil -- Password for Basic auth. Falls back to OPENCODE_SERVER_PASSWORD env var
```
mikew, line 210: Decided to add support for a function because my opencode server key is kept in a file, and I figured some neovim users would appreciate a more lazy approach to loading opencode credentials.

lua/opencode/opencode_server.lua
```diff
@@ -272,6 +275,7 @@ function OpencodeServer:spawn(opts)
   self.mode = 'serve'
   self.job = vim.system(cmd, {
     cwd = opts.cwd,
+    env = auth.get_env(),
```
mikew, line 278: Wasn't sure if the credential settings should also extend to the server that opencode.nvim starts. Since it was so small I figured it wouldn't hurt adding it and then seeing if it's appropriate or not.

lua/opencode/auth.lua
```diff
@@ -0,0 +1,78 @@
...
+---@type {password: string | nil, username: string} | nil
+local cache = nil
+
+--- Resolve a credential value that may be a string or a function returning a string.
+--- Returns nil for nil, empty string, or function errors.
+---@param val string | (fun(): string | nil) | nil
+---@return string | nil
+local function resolve_credential(val)
+  if type(val) == 'function' then
+    local ok, result = pcall(val)
```
jensenojs, line 14: Should credential function errors be surfaced here instead of being treated as missing credentials and cached?

lua/opencode/opencode_server.lua
```diff
@@ -95,6 +96,7 @@ function OpencodeServer.health_check(url, timeout_ms)
   curl.request({
     url = url,
     method = 'GET',
+    headers = auth.get_auth_headers(),
```
jensenojs, line 99: another thing I noticed: `curl.request()` includes the Authorization header in its debug log, which may expose the Basic Auth credentials.

sudo-tee:
@mikew 

This is a nice feature to add

Thanks a lot.

Before I merge this can you add the new config fields to the README ?

A part from that great job :)

mikew:
Absolutely! Will get those in a bit after I've had some coffee

mikew:
Sorry that took so long!

sudo-tee:
Thanks a lot, 

This is a really great 1st contribution :)

lua/opencode/opencode_server.lua
sudo-tee: Yes this could be an issue. I will have a look

lua/opencode/auth.lua
sudo-tee: Although not major. This could be a nice to have

lua/opencode/opencode_server.lua
sudo-tee: https://github.com/sudo-tee/opencode.nvim/issues/467

I created an issue so I don't forget to have a look

PR #466, opened 2026-07-13
feat(ui): add multi-select support for dialog questions

sudo-tee:
This MR implements the question with multiple choices.

image
image

It also supports entering a custom option.


sudo-tee:
@jensenojs 

You might be interested in this functionality.

Let me know what you think

jensenojs:
I do have some comments, and I will submit a patch later. Additionally, I noticed that you ignored some of my comments on other PRs; I wanted to confirm whether you missed them or if you felt those comments were just minor issues that could be overlooked (｡･ω･｡)ﾉ♡

jensenojs:
I tested this locally and put together a patch. 

What I found

- Single- and multi-select questions need different submission semantics. A single choice may complete immediately, while multi-select options and custom input should only update a draft. An explicit Confirm action should submit it.
  - Often, the point of an "Other" option in multiple-choice questions (at least for me) is to supplement the existing choice with additional information
- Mixed single/multiple requests must derive their interaction mode from the current question rather than retaining the first question’s Dialog behavior.
- The protocol boundaries should be represented directly: respect custom=false, allow an empty multi-select answer [], and prevent callbacks from an old question from modifying a newer request.


The diagram below illustrates what I think might be a reasonable interaction design, with the hope of reducing your review effort

```
Question
├── Single select — choosing completes the answer
│   ├── Regular option
│   │   └── Enter / Tab / number → submit
│   │
│   └── Other (when custom is allowed)
│       ├── Open the input editor
│       └── Input Enter → submit
│
└── Multiple select — edit a draft, then confirm
    ├── Regular option
    │   └── Enter / Tab / Space / number → toggle draft selection
    │
    ├── Other (when custom is allowed)
    │   ├── Open the input editor
    │   └── Input Enter → save the draft and return without submitting
    │
    └── Confirm
        ├── Enter is the only submission key
        ├── Non-empty draft → submit selected answers
        └── Empty draft
            ├── First Enter → show an empty-answer confirmation
            └── Second Enter → submit []
Switch question
└── Rebuild the interaction from the current question's
    single / multiple / custom fields

Async callback
└── May only modify its own request_id + question_index
```

double check, sth like
Pasted Graphic 2


patch here, @sudo-tee 

[pr-466-local-fixes.patch](https://github.com/user-attachments/files/29998649/pr-466-local-fixes.patch)

sudo-tee:
> I do have some comments, and I will submit a patch later. Additionally, I noticed that you ignored some of my comments on other PRs; I wanted to confirm whether you missed them or if you felt those comments were just minor issues that could be overlooked (｡･ω･｡)ﾉ♡

Thanks.

If I missed some comments or anything I am sorry . Could you point me to theses questions  you are referring to ?

sudo-tee:
> I tested this locally and put together a patch.
> 
> What I found
> 
> * Single- and multi-select questions need different submission semantics. A single choice may complete immediately, while multi-select options and custom input should only update a draft. An explicit Confirm action should submit it.
>   
>   * Often, the point of an "Other" option in multiple-choice questions (at least for me) is to supplement the existing choice with additional information
> * Mixed single/multiple requests must derive their interaction mode from the current question rather than retaining the first question’s Dialog behavior.
> * The protocol boundaries should be represented directly: respect custom=false, allow an empty multi-select answer [], and prevent callbacks from an old question from modifying a newer request.
> 
> The diagram below illustrates what I think might be a reasonable interaction design, with the hope of reducing your review effort
> 
> ```
> Question
> ├── Single select — choosing completes the answer
> │   ├── Regular option
> │   │   └── Enter / Tab / number → submit
> │   │
> │   └── Other (when custom is allowed)
> │       ├── Open the input editor
> │       └── Input Enter → submit
> │
> └── Multiple select — edit a draft, then confirm
>     ├── Regular option
>     │   └── Enter / Tab / Space / number → toggle draft selection
>     │
>     ├── Other (when custom is allowed)
>     │   ├── Open the input editor
>     │   └── Input Enter → save the draft and return without submitting
>     │
>     └── Confirm
>         ├── Enter is the only submission key
>         ├── Non-empty draft → submit selected answers
>         └── Empty draft
>             ├── First Enter → show an empty-answer confirmation
>             └── Second Enter → submit []
> Switch question
> └── Rebuild the interaction from the current question's
>     single / multiple / custom fields
> 
> Async callback
> └── May only modify its own request_id + question_index
> ```
> 
> double check, sth like Pasted Graphic 2
> 
> patch here, @sudo-tee
> 
> [pr-466-local-fixes.patch](https://github.com/user-attachments/files/29998649/pr-466-local-fixes.patch)

I will  have a look at your patch later today, Thanks for putting it together

jensenojs:
- https://github.com/sudo-tee/opencode.nvim/pull/465
- https://github.com/sudo-tee/opencode.nvim/pull/461

No worries, I just assigned an agent to conduct some reviews; there shouldn't be any major issues

> > I do have some comments, and I will submit a patch later. Additionally, I noticed that you ignored some of my comments on other PRs; I wanted to confirm whether you missed them or if you felt those comments were just minor issues that could be overlooked (｡･ω･｡)ﾉ♡
> 
> Thanks.
> 
> If I missed some comments or anything I am sorry . Could you point me to theses questions you are referring to ?

sudo-tee:
I don't see any comments by you on the 2 linked PR.

jensenojs:
oh, I forgot I submitted it—maybe it's available to view now?

sudo-tee:
> oh, I forgot I submitted it—maybe it's available to view now? 

Answered them.

sudo-tee:
@jensenojs 

I quite like the updates you did. Thanks

There is 1 thing I am not sure. In a previous comment with the inline Input you said that "dismissing" the inline input by accident (changing cursor) you would lose the content of the input. With this patch the "issue" is back,

Is that the intended behavior ?

jensenojs:
> @jensenojs
> 
> I quite like the updates you did. Thanks
> 
> There is 1 thing I am not sure. In a previous comment with the inline Input you said that "dismissing" the inline input by accident (changing cursor) you would lose the content of the input. With this patch the "issue" is back,
> 
> Is that the intended behavior ?

Ah, that doesn't sound right. It must be some oversight, thk @sudo-tee 

--

update + 1

To be honest, and sorry that I didn't spend much time on this vibe (because I've been a bit busy lately), so I just sent the patch directly to you. Feel free to adjust it

---

update + 2

For the inline other option, I think data ownership may need some adjustment. For example, the buffer draft should probably belong to the 'question,' since we might reuse this mechanism elsewhere. Also, when a new question comes in, the old cached text becomes useless, so we'll need some cleanup or invalidation logic. I do have a local branch for it, but I don't really have the bandwidth to dive into it right now."

sudo-tee:
No worries, 

This is a great idea to move the "state" to the question.  I will play with it a little bit

jensenojs:
Feel free to @ me if you have any other review needs; (｡･ω･｡)ﾉ♡

PR #468, opened 2026-07-15
fix(ui): stabilize custom question input and dialog restoration

jensenojs:
## Summary

iShot_2026-07-15_19 17 54

I noticed that the floating window doesn't wrap text automatically, and <c-c> doesn't seem to save drafts. Here are some adjustments I've made; I'd love to hear your thoughts on them.

---

```text
Inline Custom Answer
├── Choose a fixed width when opened
│   ├── Up to 50 columns
│   └── Shrink near the right edge
├── Wrap input without changing width
├── Resize vertically with multiline content
│   └── Stop growing at the bottom edge
└── Preserve the complete draft
    ├── Submit long and multibyte text without truncation
    └── Restore multiline drafts after cancellation
```

I also did some additional code cleanup and bug fixes

- Restore the active question dialog after panel layout changes.
- Centralize the duplicated Other-input lifecycle and keep its tests on real scheduling semantics.

### DIFF
--- a/lua/opencode/ui/inline_input.lua
+++ b/lua/opencode/ui/inline_input.lua
@@ -19,14 +19,15 @@ function M.open(opts)
   local anchor = vim.fn.screenpos(opts.win, opts.row + 1, opts.col + 1)
   local width = math.max(1, math.min(50, vim.o.columns - anchor.col - 1))
   local max_height = math.max(1, vim.o.lines - anchor.row - 2)
+  local initial_lines = opts.initial_text and vim.split(opts.initial_text, '\n', { plain = true }) or nil
 
   local buf = vim.api.nvim_create_buf(false, true)
   vim.bo[buf].buftype = 'prompt'
   vim.bo[buf].bufhidden = 'wipe'
   vim.fn.prompt_setprompt(buf, '')
 
-  if opts.initial_text then
-    vim.api.nvim_buf_set_lines(buf, 0, 1, false, { opts.initial_text })
+  if initial_lines then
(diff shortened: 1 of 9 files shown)


jensenojs:
I’m using this model to keep draft ownership and panel restoration consistent throughout the implementation. I've recently taken a liking to "reviewing" code this way

```
question_window
│
├─ current question / current index
├─ collected answers / multi selections
└─ per-question Other draft
         │
         │ open edit
         ▼
inline_input
├─ transient prompt buffer
├─ transient float window
├─ fixed width + wrapped height
└─ submit / cancel-with-full-draft
         │
         ├─ submit ──> question commits answer and clears draft
         └─ cancel ──> question stores draft and redraws dialog

Panel lifecycle
├─ hidden-buffer restore ──> reuse existing buffer and Dialog state
└─ fresh reopen ───────────> renderer rebuilds output, then restores pending question
```

jensenojs:
GPT generated several corner cases, which have been fixed together in this commit

lua/opencode/ui/inline_input.lua
```diff
@@ -18,37 +16,65 @@ local M = {}
...
     height = 1,
     style = 'minimal',
     border = 'rounded',
     title = opts.title and (' ' .. opts.title .. ' ') or nil,
     title_pos = opts.title and 'left' or nil,
     zindex = 60,
   })
+  vim.wo[win].wrap = true
+
+  local function resize_height()
```
sudo-tee, line 51: Really nice improvement to auto resize

sudo-tee:
Nice cleanup. Thanks for the PR

PR #470, opened 2026-07-15
fix(ui): preserve mounted layout across position changes

jensenojs:
## Summary

```text
Layout position
├── config.ui.position    -> where the next windows should open
└── windows.position      -> where the mounted windows actually opened
    ├── close/hide uses the mounted position
    └── restore/new windows use the current config
```

This prevents a `current` layout from closing its code window when configuration changes to `right` before teardown finishes.

- Sequence both toggles in the three-state layout recipe before changing position.
- Preserve mounted position through navigation, sizing, hiding, and teardown.
- Add regression coverage for current-to-right transitions and hidden restoration.

## Verification

- `./run_tests.sh -t tests/unit/persist_state_spec.lua`
- `./run_tests.sh -t tests/unit/navigation_spec.lua`
- `./run_tests.sh -t tests/unit/zoom_spec.lua`
- `./run_tests.sh`

### DIFF
--- a/docs/recipes/three-state-layout/demo.lua
+++ b/docs/recipes/three-state-layout/demo.lua
@@ -26,32 +26,34 @@ local ACTIONS = {
     if not config then
       return
     end
-    api.toggle(false)
-    config.ui.position = 'right'
-    api.toggle(false)
+    return api.toggle(false):and_then(function()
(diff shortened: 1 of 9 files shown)


sudo-tee:
Thanks for the PR, seems like a reasonnable fix

PR #471, opened 2026-07-16
Feat/451 chorefeature tell opencode what to do differently

### LINKED ISSUE #451: [chore/feature] : Tell OpenCode what to do differently

sudo-tee:
This pr adds the possibility to `Tell opencode` whay you deny a permission. 

This step is important because it does not interrupt subagents and they adapt to the comment you give.

image

To deny the request and stop you can use the the new Double `esc`

### DIFF
--- a/lua/opencode/api_client.lua
+++ b/lua/opencode/api_client.lua
@@ -385,7 +385,7 @@ end
 --- Respond to a permission request
 --- @param id string Session ID (required)
 --- @param permissionID string Permission ID (required)
---- @param response_data {response: "once"|"always"|"reject"} Response data
+--- @param response_data {response: "once"|"always"|"reject", message?: string} Response data
 --- @param directory string|nil Directory path
 --- @return Promise<boolean>
 function OpencodeApiClient:respond_to_permission(id, permissionID, response_data, directory)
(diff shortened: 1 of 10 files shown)


jensenojs:
~~sorry for being late, i will test and review now~~

---

here are some comment about life cycle

```
fix(permissions): bind feedback UI to request lifecycle

permission request              # authoritative state
└── interaction                 # scoped to request ID
    ├── feedback input
    ├── double-Esc timer
    ├── processing lock
    └── Dialog-owned Esc mapping

Removing a permission now closes its input, stops its timer, invalidates stale
callbacks, and removes its keymap. Responses are claimed once, cancelled input
leaves the request pending, and session cancellation uses the current reply
payload.
```

[pr-466-local-fixes.patch](https://github.com/user-attachments/files/30148299/pr-466-local-fixes.patch)

---

Actually, it might be nice if we could press 'S' to jump to the subagent's context when the subagent pops up this request. I may create another patch later. As always, you’re free to use or modify any of the patches as you see fit (Because to be honest, I haven't been paying much attention to these patches because of  other matters)

图片

What do you think? We're starting to (or have already) developed interactions that go beyond TUI.

图片

[pr-471-subagent-permission-session-action.patch](https://github.com/user-attachments/files/30148756/pr-471-subagent-permission-session-action.patch)


@sudo-tee

jensenojs:
Furthermore, just thinking ahead—perhaps once the UI interactions have stabilized, we could use an agent to re-record all the interaction GIFs and give our README a fresh makeover.

Of course, this is still a long way off, because even getting an agent to record is quite a hassle, let alone the fact that none of these tasks have been completed yet

sudo-tee:
@jensenojs I think the 1st patch is not the right one. The is the ole one I applied for another MR.

jensenojs:
[pr-471-permission-lifecycle-fix.patch](https://github.com/user-attachments/files/30184571/pr-471-permission-lifecycle-fix.patch)

ah yes, this one @sudo-tee

sudo-tee:
I've applied the patches,

I like it so far :) thanks. 

I like the idea of creating new gifs from an agent :).  IT might be tricky to get it right though.

jensenojs:
> I like the idea of creating new gifs from an agent :). IT might be tricky to get it right though.

When that day finally comes, I'll handle the workflow for automatically generating GIFs.

PR #472, opened 2026-07-18
docs: update README with configuration for blink.cmp

NeoPrint3D:
This documentation update includes guidance for preventing unwanted completions caused by context pollution in OpenCode prompt windows when using standard NeoVim configurations such as LazyVim. It helps users avoid generic Blink.cmp suggestions such as buffer words, dates, and other irrelevant entries, that can interrupt prompts, cause accidental completions, and obscure OpenCode’s intended file-mention and command suggestions.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,9 +110,6 @@ Install the plugin with your favorite package manager. See the [Configuration](#
 }
 ```
 
-
-
-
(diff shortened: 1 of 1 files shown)


jensenojs:
Thank you for your contribution; let me see if I can make it better

PR #439, opened 2026-07-02
[feat] : Add inline message actions

jensenojs:
Reverting the latest message in the TUI is a very common action for me; I fine-tune my prompts by observing the model's output

iShot_2026-07-02_17 27 16_under10M

PR #473, opened 2026-07-20
[feat]: Add contextual actions to user messages

jensenojs:
## Summary

- expose Revert, Copy, and Fork through the existing contextual actions UI
- keep one action set across the full rendered user message range
- restore overwritten buffer-local mappings when contextual actions become inactive
- copy original non-synthetic user text through the existing API command axis

## Model

```text
state.messages                         RenderState
├── message id ──> undo / fork         └── user message range
└── original text ──> copy                 └── R / C / F

cursor line ──> actions in range ──> temporary mappings ──> opencode.api
                       │
                       └── leave range ──> restore previous mappings
```

## Testing

- `./run_tests.sh`
- 138 replay cases
- contextual action mapping, persistent-buffer, range-shift, and Copy contract coverage

Closes the interaction gap discussed in #439 by reusing contextual actions instead of adding another dialog or mouse-specific surface.

### DIFF
--- a/lua/opencode/api.lua
+++ b/lua/opencode/api.lua
@@ -52,6 +52,7 @@ local action_groups = {
     open_input_new_session_with_title = session.open_input_new_session_with_title,
     rename_session = session.rename_session,
     undo = session.undo,
+    copy_message = session.copy_message,
     fork_session = session.fork_session,
     toggle_session_lock = session.toggle_session_lock,
   },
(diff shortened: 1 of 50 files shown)


jensenojs:
test for copy

iShot_2026-07-20_19 13 11

---

test for revert

iShot_2026-07-20_19 13 50

---

I also tested the fork(but I'm too lazy to record another GIF); this feature might make you more inclined to try adjusting your prompts by reverting the previous message @sudo-tee

sudo-tee:
This is great.

I much prefer this version with contextual actions on the user message than the previous one :)

Thanks

PR #477, opened 2026-07-26
feat(renderer): skip reasoning in next/prev message navigation

phanen:
Most of time I want to skip reasoning after jump.


sudo-tee:
This is a good idea.

Thanks for the PR. seems to work well

PR #479, opened 2026-07-26
feat: remove winfixbuf and make ctrl w o work in opencode window

jugarpeupv:
Remove winfixbuf from output/input windows so that file pickers (telescope, snacks, fyler) and :e can switch the buffer without E1513 errors. Extend the BufEnter teardown guard to all positions.

When the buffer is replaced externally, keep the output window open with the new file buffer instead of closing it, so the file remains visible after opencode UI tears down.

## Summary

When `winfixbuf=true` was set on opencode output/input windows (split and float positions), any attempt to switch the buffer via file pickers (telescope, snacks, fyler) or `:e` would throw `E1513: Cannot switch buffer. 'winfixbuf' is enabled` — a hard error with no graceful recovery.

## Solution

Three coordinated changes:

1. **Remove `winfixbuf=true`** from output and input windows — lets external tools switch the buffer without E1513.
2. **Extend `BufEnter` teardown guard to all positions** — was previously conditional on `position == 'current'`. Now detects when a non-opencode buffer enters an opencode window and tears down the UI cleanly.
3. **Keep the output window open when the buffer was replaced externally** — `close_or_restore_output_window` now checks if the buffer already changed before closing the window. If it was replaced by fyler/telescope/etc., the window stays open with the new file buffer visible.

## Files changed

| File | Change |
|------|--------|
|  `lua/opencode/ui/output_window.lua` | Removed `winfixbuf=true` for non-`current` positions |
|  `lua/opencode/ui/input_window.lua` | Removed `winfixbuf=true` for non-`current` positions |
|  `lua/opencode/ui/autocmds.lua` | Made `BufEnter` teardown guard apply to all positions (was `current`-only) |
|  `lua/opencode/ui/ui.lua` | `close_or_restore_output_window`: detect buffer replacement, keep window open |

## Behaviour

| Scenario | Before | After |
|----------|--------|-------|
| `:e file` in opencode window | `E1513` error | Opens file, opencode UI tears down cleanly |
| Telescope/snacks pick file from opencode window | `E1513` error | Opens file in window, opencode UI tears down |
| Fyler open file targeting opencode window | `E1513` error | Opens file in window, keeps window open, tears down rest |
| Normal opencode toggle off | Window closes (unchanged) | Window closes (unchanged) |


jugarpeupv:
> The `inline_input restores multiline initial_text and places the cursor at its end` test failed on `macos-latest, v0.11.4` — it's a pre-existing flaky test unrelated to these changes. The `vim.wait(50, ...)` at  `tests/unit/inline_input_spec.lua:147` times out when cursor position doesn't settle within 50ms on macOS CI. Bumped the timeout to 100ms to make it more resilient.

jugarpeupv:
This does not work well either X.X

PR #483, opened 2026-08-13
feat(ui): add <C-p>/<C-n> history navigation to question dialog input

phanen:
Problem: the 'Type your own answer' inline input box did not inherit
any input history, so users could not scroll back through prior answers
the way `vim.fn.input()` allows with `<C-p>`/`<C-n>`.

Solution: bind insert-mode `<C-p>`/`<C-n>` in `inline_input.lua` that
walk vim's built-in `input` history via `vim.fn.histget`/`histadd`,
sharing the same history list as `vim.fn.input()` itself.


sudo-tee:
Thanks for the PR.

This is a great addition

PR #489, opened 2026-09-10
fix: widen symbol target candidate sources in output window

jensenojs:
Symbol jump in the output window often missed because the candidate file set was too narrow: files @-mentioned in user messages were dropped, bare-name mentions failed the cwd-relative readability check, and the formatter only accepted files mentioned before the token.

Now `reference_facts.available_files()` unions readable ref files with loaded plain buffers, user file parts enter reference facts, the position narrowing is gone, and navigation re-reads the set at keypress time. `gd` is also bound to `jump_to_target_at_cursor` in the output window.

Full suite passes; added an e2e regression covering open-buffer-as-only-candidate.


jensenojs:
I'm back! @sudo-tee

sudo-tee:
Thanks for the PR, this looks like a nice fix

Issue #487, opened 2026-09-07
Feature request: support OpenCode v2 API/protocol


wzj-zz:
### Feature request

Please add support for the OpenCode v2 API/protocol while keeping existing OpenCode v1 users working.

`opencode.nvim` currently targets the v1 server API. For example, the client calls `/session/{id}/message`, `/session/{id}/abort`, `/global/health`, and subscribes to `/global/event` (see `lua/opencode/api_client.lua` and `lua/opencode/opencode_server.lua`). OpenCode v2 exposes a different HTTP surface under `/api`, including `/api/health`, `/api/event`, `/api/session/{id}/prompt`, `/api/session/{id}/interrupt`, `/api/session/{id}/message`, and location/workspace-aware request and response models.

The current plugin therefore cannot be used with an OpenCode v2 server, even though the v2 API is now documented and available here:

- https://opencode.ai/v2/docs/api
- https://opencode.ai/v2/openapi.json

### Suggested scope

- Detect the server/API version at startup, or provide an explicit `api_version = "v1" | "v2"` option.
- Add a v2 API client/adapter for the core workflows: health checks, session create/list/load/delete/rename, send prompt, abort/interrupt, message/history loading, permissions, questions/forms, agents/models, file/context lookup, and MCP.
- Subscribe to the v2 event stream and normalize v2 events into the existing event manager/renderer model. The v2 event payloads include versioned durable sync events and location/workspace metadata, so directory/workspace filtering should be preserved.
- Map v2 response shapes to the existing UI state where possible, while exposing v2-only concepts such as workspaces, cursor pagination, session inbox/queue, and forms incrementally.
- Update the documented minimum supported OpenCode version and add replay/unit coverage for representative v2 responses and events.

### Expected behavior

- With OpenCode v1, current behavior remains unchanged.
- With OpenCode v2, opening a session and submitting a prompt should work without manually patching the plugin or running a compatibility proxy.
- Unsupported v2-only features should fail clearly rather than silently leaving the UI stale.

I am happy to help test this against the v2 server and provide sample event/API payloads if useful.

sudo-tee:
Thanks for the report.

This might require a hefty refactor to achieve this. I will have a look at this in the near future

jensenojs:
It would be great if you could help with the testing (｡･ω･｡)ﾉ♡

sudo-tee:
Of course I will test this.

sudo-tee:
@wzj-zz 

[...]

Issue #476, opened 2026-07-25
Feature request: render different session in separate buffer


phanen:
I often navigate between different session in single nvim instance. But it feels laggy since all session need to be re-render in a single buffer, also the cursor position is reset.

ALso if we render different in separate buffer, maybe cursor position can be remember automatically `:h 'quote`.

sudo-tee:
@phanen  

I am not against the idea, it could be useful.

But I am not sure what would be the workflow. do you want to interact with the other session as well ? or just want to view the content of another session ?

Would a session viewer in a floating pane works ?  or another split ? or Tabs in the output window ?

phanen:
> But I am not sure what would be the workflow. do you want to interact with the other session as well ? or just want to view the content of another session ?

I navigate and create new sessions frequently. To ask them do different thing at the same time or ask different question.
```lua
          ['<a-p>'] = { 'navigate_session_tree', 'backward', mode = { 'n', 'i' } },
          ['<a-n>'] = { 'navigate_session_tree', 'forward', mode = { 'n', 'i' } },
          ['<a-,>'] = { 'open_input_new_session', mode = { 'n', 'i' } },
```

Each time I use only one session, but just switch session frequently. This operation will redraw in a single buffer from scratch and reset cursor position. Each session can be assigned to a scratch buffer (probably with session id as unique bufname so we can find/resue it again without maintaining state.

> Would a session viewer in a floating pane works ? or another split ? or Tabs in the output window ?

I'm satisfied with the current layout, but that can be discussed later. If we need to render multiple sessions at the same time (I'm not sure how it would be looks like, maybe user can toggle different session like `:Opencode toggle <session_id>`), but it inevitably means that each session should have its own buffer.

sudo-tee:
Ok I get it now. 

I will work something out. It might take a bit as I will soon be on vacation for the comming weeks.

sudo-tee:
Now that opencode support tabs natively. I might got this route

phanen:
I've been used another workaround recently, to disable treesitter completely, which make thing way much faster. (: treesitter is always the bottleneck

sudo-tee:
Should be fixed by the new tabs feature.

PR #490, opened 2026-09-10
Wip/session tabs

sudo-tee:
Adds logical session tabs inside the Opencode panel without creating Neovim tabpages.

- Each tab preserves its own session, input, output, context, model, and renderer state.
- Adds tab-strip navigation, index selection, overflow handling, pending prompt markers, and background notifications.
- Keeps tab output synchronized while sessions run in the background.
- Supports opening child-session and fork actions in new panel tabs.
- Hides tab strip when only one session tab exists.

## Configuration

Inline child-session and fork actions remain default behavior. To open them in new panel tabs:

```lua
require('opencode').setup({
  ui = {
    output = {
      actions = {
        open_in_new_tab = true,
      },
    },
  },
})
```

image

image


sudo-tee:
This is an experimental feature that I am working on since a couple of weeks and whant to gather feedback, I really like the workflow on my part, especially for navigating between sessions

@jensenojs 

@phanen I know you wanted a similar feature in #476 

Let me know if you like it or not :)

jensenojs:
I'll try it tomorrow morning (｡･ω･｡)ﾉ♡

jensenojs:
~~I'm not entirely sure if `<leader>o</>` is a good idea (though `<leader>o h/l` is fine). On my local machine, CSI-u seems to be escaping it into something strange.~~

ignore, my stupid mistakes

sudo-tee:
But I still like your idea of  <leader>o h/l and even adding <leader>o left/right might work. I will have a look

jensenojs:
> But I still like your idea of o h/l and even adding o left/right might work. I will have a look

or <leader>o, / <leader>o. (I was being stupid today because I was clicking this one lol, but it doesn't matter)

jensenojs:
I just wanted to remind you that we might need to distinguish between wanting to split different sessions within a single session-window (which is obviously different from splitting different session-windows). I currently have this bound to <keymap_prefix>h/l, though I will likely bind this lovely shortcut to your feature instead. We just might need to avoid confusing newcomers in the documentation, but it's only a minor issue.

Well, maybe not entirely, because when I switch between different sessions within a session window, the original title disappears—though, of course, that might be the expected behavior

图片

sudo-tee:
Yes this is an intended behavior, since the "session" is being replaced in the current buffer, so it makes sense that the title changes accordingly.

But I can keep the current keymaps, since users can change them anyway if they want.

jensenojs:
图片

---

opencode-v2 now has a similar feature as well, and its architecture appears to natively support two-way synchronization between TUI and Nvim.

The only thing that's kind of annoying is that the APIs seem to be all messed up. I've been trying to find time to look into this over the past few days so we can make some adjustments ahead of time (though it's not that far in advance—it should be released soon).

sudo-tee:
I also started to look a bit into opencodev2, but it looks like a massive change. 

I will need to come up with plan, as soon a lot of users will ask for compatibility.

There is already an issue opened about it there -> #487

jensenojs:
I created a branch for local emergency use. I can also participate in the design and review later; it just so happens that I've only upgraded opencode2 on one machine, so I can test the compatibility.

Although you uploaded your spec file by mistake, I still think this design document could exist in a draft branch, as it might be more helpful for future design communication

sudo-tee:
I will merge this, ans start looking into opencode2 this week

jensenojs:
I have also locally set a goal to create a "simple enough" goal for codex (which has undergone considerable end-to-end testing), but I haven't used it very deeply in opencode2 and opencode1 yet. I will try my best to bring it up this week for your reference

sudo-tee:
@jensenojs  Yes publish what you have , it might serves as a base.

PR #492, opened 2026-09-11
perf: cache parsed symbol snapshots to avoid redundant tree-sitter parses

sudo-tee:
I was having performances issues with large sessions.


sudo-tee:
@jensenojs I want to bring your attention here before the tabs.

I was having performance issues with the tokenization on long session, therefore I implemented a LRU cache for it. 

Since I don;t know exactly how you use the feature can you let me know if this works as intendend for you ?

jensenojs:
Tree-sitter's parsing scope shouldn't be very large unless there's a leak. I'll take a closer look when I get home later.

jensenojs:
Because it is still generally lazy-loaded, the file-list mentioned in the session and the actual Buffers currently open in the session (I'm ashamed to admit I only encountered this issue two days ago, but I've been using only codex for a long time recently) need to be parsed.

It feels like there are at most a dozen or so files? Most of the time? I'm mainly unsure if adding an LRU itself can fundamentally solve this problem. If it can, then implementing an LRU for the whole thing would be great. Based on the buffer we are currently focusing on and the most recently mentioned history.

sudo-tee:
I'm not sure I understand your comment

sudo-tee:
I did some profiling with `test/data/perf.json`

image

jensenojs:
> I'm not sure I understand your comment

I didn't follow the problem before, this is a great solution

Additionally, I noticed that you submitted the spec document earlier. Was that intentional or accidental? Regarding today's Tree-sitter feature, I'm wondering if it's necessary to translate my previous spec into English and upload it. What are your thoughts?

sudo-tee:
No the spec was an error on my part. I find putting spec document to mislead ai agents as most of the time the implementation will diverge from the original document. I will remove it.

PR #494, opened 2026-09-15
Dual-protocol client: OpenCode V1 and V2 wire APIs

jensenojs:
## Summary

Reworks the client to speak both OpenCode wire protocols (V1 1.18.x and V2 2.0.x) through a single connection/observation architecture.

## What changed

- New `protocols/{v1,v2}` modules own native endpoints and event interpretation per protocol; `ui/` consumes one read-only per-session Observation instead of reducing protocol events.
- Connection selection: authenticated health probe picks the protocol once per connection; reconnect on identity change, no per-request branching.
- V2 uses the native shared service via CLI discovery (`service status` / `get password`); V1 keeps local spawn, explicit URLs, and port coordination.
- Removed `api_client`, `event_manager`, `session`, `renderer/events`, `event_scope`, `session_scope`.
- Topbar cost/context stats now derive from observation usage (V2 `session.usage.updated` / snapshot first, latest entry fallback for V1).
- `run_tests.sh` now names the spec file when a test file dies at load time (e.g. requiring a removed module), instead of a bare non-zero exit.

## Contract

The published OpenAPI spec and the running 2.0.x server disagree at several endpoints (`/api/project/current`, `rename` via POST, prompt `command` field, `fork` body shape). This implementation follows the running server; fixtures under `tests/data/v2/` are live captures from a real 2.0.1 server. Known future break: upstream dev renames `permission.asked` / `form.*` events.

Details in `docs/drafts/v2-migration-draft.md` (temporary draft intended to be folded into docs later).

## Testing

- `./run_tests.sh` green (minimal, unit, replay).
- Per-protocol contract suites: `protocol_{v1,v2}_{operations,observation}*_spec.lua`.
- Live dual-client check done against a real 2.0.3 shared service (native discovery, auth, event stream, prompt admission, both TUI↔nvim visibility).


jensenojs:
This might be my last chance to clean up the code; I really wish I had enough time...

- https://github.com/sudo-tee/opencode.nvim/issues/322

sudo-tee:
I started to play with it. 

I just fixed a bug in v1 where the panel is in perpetual loading.

I will try to fix the issue I see as they come. 

This is a significant change from the prior architecture. I would have liked a simpler and iterative approach. There is a lot of issues at the moment so It's hard to assess if this is a good base or not.  I will try to fix some issues and see.

jensenojs:
I understand

> I started to play with it.
> 
> I just fixed a bug in v1 where the panel is in perpetual loading.
> 
> I will try to fix the issue I see as they come.
> 
> This is a significant change from the prior architecture. I would have liked a simpler and iterative approach. There is a lot of issues at the moment so It's hard to assess if this is a good base or not. I will try to fix some issues and see.

sudo-tee:
It's not bad or anything, don't get me wrong.

jensenojs:
I completely understand; even my testing on v2 was insufficient. I just urgently need a version that works on v2

besides, having a flawed implementation to critique is also very good.

sudo-tee:
It's getting there bits by bits. I am now focusing on simplifying as much as I can. 

I fixed a lot of rendering. scrolling and context management.

sudo-tee:
I did a lot of progress and cleanup. 

I was even able to gertthe topology reports all fixed `0 cycles, 0 violations, clean → /home/francis/Projects/_nvim/opencode.nvim/dependency-graph.html` while simplifying the structure.

There is still more testing to do in both v1 and v2.. but it's getting there

lua/opencode/protocols/v2/operations.lua
```diff
@@ -0,0 +1,704 @@
...
+          error('V2 submit received invalid file bytes')
+        end
+        uri = 'data:' .. file.media_type .. ';base64,' .. vim.base64.encode(file.bytes)
+      elseif type(file.server_uri) == 'string' and file.server_uri:match('^file:///') then
+        local path = file.server_uri:sub(8)
+        uri = 'file://' .. (type(path_map) == 'function' and path_map(path) or path)
+      else
+        error('V2 submit server_uri must be an absolute file URI')
+      end
+      body.files[#body.files + 1] = { uri = uri, name = file.name, mention = mention(file.mention) }
```
bivashy, line 451: that may lead to error:
```lua
[2026-09-20 14:43:14] [opencode] [ERROR] [/myhomedir/opencode.nvim/lua/opencode/services/messaging.lua:196] Error sending message to session: ...e/opencode.nvim/lua/opencode/protocols/v2/operations.lua:464: attempt to get length of field 'files' (a nil value)
```
_(ignore that stacktrace shows 464 line, it is because I've added some debug statements)_

Because in line 409:
```
body.files = nil
```

---

error appeared when i was testing that PR locally, idk if i've properly fixed it, i just needed `opencode.nvim` working, my patch: https://github.com/jensenojs/opencode.nvim/commit/617ebf85f10d63a70d0885ff3a2e6f6fa6d30c95

sudo-tee:
@jensenojs 

I think I ran out of bugs or thinks to fix/simplify.  It seems to work pretty well for me on both v1 and v2 and the code seems to be well seperated after some polishing

There ia a conflict with main I can't fix since I cannot force push on your branch.

Do you have any remaining issues with this branch ?

Before merging I will create a `v1` branch from main so people cans still refers to it if needed.

jensenojs:
Our code no longer has any circular dependencies!

sudo-tee:
Yes I used the opportuny to cleanup some cruft and even implemented a type check .
 
 My only fear is that as you saw opencode v2 still pushes breaking changes in minor versions which is annoying.`

jensenojs:
At least it won't fail silently now; there will be a smoke test before startup to remind users to upgrade either OpenCore itself or the plugins. ( https://github.com/sudo-tee/opencode.nvim/pull/494/commits/ea83ac0d86945f63baf0f109520f45578380b67c )

It is also for this reason that I am not quite sure whether it should be merged into main right now; we could just add a note in the README stating that this branch can be used for v2 compatibility

sudo-tee:
@jensenojs 
Since this PR is not a branch per say, I will create a new branch named `v2` that I will then ilnk from the main branch.

sudo-tee:
https://github.com/sudo-tee/opencode.nvim/pull/499

I will close this PR in favor of the v2 branch

PR #497, opened 2026-09-21
fix(ui): clear input contents on submit

disrupted:
Fixes regression introduced by #463

`state.input_content` kept the submitted message and subsequent window toggles kept showing it via `recover_input()`


sudo-tee:
Thanks for the fix.  I haven;t encountered this issue, and  the fix is simple enough

Thanks

